PR #20, opened 2020-11-09
Small changes to clarify the example

laurensslats:
#### Summary
People are likely to copy-paste the example vendor and change the fields accordingly.

To support them in this process, I've added the descriptions in the codec and profile files

#### Changes
- Added descriptions in the example codec and profile


johanstokking (approved):
Thanks

PR #21, opened 2020-11-18
dl-iam: set datasheet url to static location

decentlab:
#### Summary

Set DL-IAM datasheet URL to a static location. Otherwise, the current URL is a dynamically generated and may be removed any time.

#### Changes

- vendor/decentlab/dl-iam.yaml: changed dataSheetURL field.


johanstokking (approved):

johanstokking:
Thanks!

PR #25, opened 2020-11-23
Fix Profile IDs for The Things Node to match filenames

neoaggelos:
#### Summary

Quickfix PR for an issue that came up while working on https://github.com/TheThingsNetwork/lorawan-stack/issues/3433. Profile IDs for The Things Node do not match with the existing files.

#### Changes

- Rename profile `the-things-node-eu868` to `the-things-node-868`, matching `the-things-node-868.yaml` file.
- Rename profile `the-things-node-us915` to `the-things-node-915`, matching `the-things-node-915.yaml` file.

#### Notes for Reviewers

There should probably be a CI check that will make sure that no such inconsistencies are allowed.

### DIFF
--- a/vendor/index.yaml
+++ b/vendor/index.yaml
@@ -1537,5 +1537,5 @@ vendors:
     name: Exedra General Trading and Contracting
     vendorID: 841
 
-  # - id: the-things-products
-  #   name: The Things Products
+  - id: the-things-products
+    name: The Things Products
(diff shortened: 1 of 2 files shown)


johanstokking:
> There should probably be a CI check that will make sure that no such inconsistencies are allowed.

It's caught by CI but `the-things-products` was commented out as vendor.

Fixed that with https://github.com/TheThingsNetwork/lorawan-devices/pull/25/commits/41c79095c54ab6087cc16167cdee33bb0b907274

johanstokking (approved):

PR #26, opened 2020-11-27
Corrected T-Code for Home Sensor

akashpanchal92:
#### Summary

Added correct T-Code for Home Sensor
...

#### Changes

- ...
- ...

#### Notes for Reviewers

...

#### Release Notes

- ...

### DIFF
--- a/vendor/tektelic/index.yaml
+++ b/vendor/tektelic/index.yaml
@@ -1,3 +1,3 @@
 endDevices:
-  - t000489x-home-base
-  - t00048xx-home-pir
+  - T0004893-home-base
+  - t0004885-home-pir


vendor/tektelic/index.yaml
```diff
@@ -1,3 +1,3 @@
 endDevices:
-  - t000489x-home-base
-  - t00048xx-home-pir
+  - T0004893-home-base
```
johanstokking, line 2:
```suggestion
  - t0004893-home-base
```

johanstokking: Must be lowercase alphanumeric with dashes

johanstokking (changes requested):
What is the aim of this?

I'd expect that the `x` stands for compatibility with any number.

If this is indeed wrong, then the file names should be updated too, because these identifiers point to the filename. Please see also the validation output.

johanstokking:
Please reopen as needed

PR #32, opened 2020-12-14
Changes for Home Sensor

akashpanchal92:
#### Summary

...

#### Changes

- ...
- ...

#### Notes for Reviewers

...

#### Release Notes

- ...

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -260,9 +260,9 @@
       "integrity": "sha512-k/vGaX4/Yla3WzyMCvTQOXYeIHvqOKtnqBduzTHpzpQZzAskKMhZ2K+EnBiSM9zGSoIFeMpXKxa4dYeZIQqewQ=="
     },
     "node_modules/ini": {
-      "version": "1.3.5",
-      "resolved": "https://registry.npmjs.org/ini/-/ini-1.3.5.tgz",
-      "integrity": "sha512-RZY5huIKCMRWDUqZlEi72f/lmXKMvuszcMBduliQ3nnWbx9X/ZBQO7DijMEYS9EhHBb2qacRUMtC7svLwe0lcw==",
+      "version": "1.3.8",
+      "resolved": "https://registry.npmjs.org/ini/-/ini-1.3.8.tgz",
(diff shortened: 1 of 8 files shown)


johanstokking (changes requested):
@akashpanchal92 thanks for your contribution!

The continuous integration failed because of validation errors. Here are a few tips;

1. The codecs need to be in the new format for The Things Stack. You reverted them to the original The Things Network Stack V2. This doesn't work with the device repository
2. We don't support modules; only ECMAScript 5.1
3. The regional parameters version per region is always one version. Also, if the MAC version is 1.0.3, the regional parameters version will be 1.0.3-A. If the MAC version is 1.0.2, the regional parameters version can be 1.0.2 or 1.0.2-revB

akashpanchal92:
Hi @johanstokking : Thank you for your comments. I could understand third point mentioned in your comment regarding alignment of MAC version and regional parameters. I tried making changes and then committed the same. However, it couldn't validated. Can you please let me know the reason behind it?

Also, I could not understand point#1 and 2 in your comment. Would you mind providing me more details?

johanstokking:
> Also, I could not understand point#1 and 2 in your comment. Would you mind providing me more details?

Regarding the two points:

1. You are using the [`function Decoder()`](https://github.com/TheThingsNetwork/lorawan-devices/pull/32/files#diff-ed93f9011026b680fb9889548b4cecd58d8acaee1391a09cdbcf2a86a38bfe74R1) declaration, while this repository requires `function decodeUplink()`. Please see the documentation: https://github.com/TheThingsNetwork/lorawan-devices#payload-codecs.
2. There's now an [module export](https://github.com/TheThingsNetwork/lorawan-devices/pull/32/files#diff-ed93f9011026b680fb9889548b4cecd58d8acaee1391a09cdbcf2a86a38bfe74R802) which is not supported

You can locally validate everything using `make validate`. First run `make deps` first.

johanstokking:
@akashpanchal92 can you address the changes requested?

akashpanchal92:
> @akashpanchal92 can you address the changes requested?

Hi @johanstokking : Apologies for the delay in getting back to you. I have added my team member, Taras, as a collaborator for my repository. We have finished making all changes and Taras is helping me fix the validation issues. We are hoping that we shall be able to close this action item by tomorrow. I sincerely apologize for the delay.

PR #35, opened 2020-12-23
Add bvs and wwd device

christianpaetz:
#### Summary

...

#### Changes

- added Aqua-Scope as new vendor
- added device description for Ball Valve Servo 'bvs'
- added device description for Water Sensor 'wwd'

#### Notes for Reviewers

...

#### Release Notes

- ...


vendor/aquascope/bvs.js
```diff
@@ -0,0 +1,52 @@
+function decodeUplink(input) {
+  var t = input.bytes[2]*0xff+input.bytes[3]
+  switch (input.fPort) {
+    case 10:
+      return {
+        // Decoded data
+        data: {
+          leak: (input.bytes[0] & 0x01) ? "Y":"N",
+          valve: (input.bytes[0] & 0x20) ? "On":"Off",
```
johanstokking, line 9: Do you prefer this over booleans?

vendor/aquascope/bvs.js
```diff
@@ -0,0 +1,52 @@
...
+      };
+    default:
+      return {
+        errors: ['unknown FPort'],
+      };
+  }
+}
+
+function encodeDownlink(input) {
+  if (input.data.valve == "On") r = 255; else r = 0;
```
johanstokking, line 29: Idem; `"on"` or `true` would not trigger the valve

johanstokking (approved):
Great stuff, thanks. Only a few questions. I understand if upstream applications depend on a particular string like `On` or `Y`, but otherwise I would prefer using booleans for that. Up to you.

Can you please sign the CLA with your GitHub account?

christianpaetz:
cla is signed. You are right, boolean is way more elegant. Will change with next FW version revision.

PR #38, opened 2021-01-07
Add the STMicroelectronics nucleo-WL55JC development boards

pierrefraboulet-st:
#### Summary

Add the STMicroelectronics nucleo-WL55JC development boards

#### Changes

- ...
- ...

#### Notes for Reviewers

...

#### Release Notes

- ...


vendor/stmicroelectronics/nucleo-wl55jc-profile-cn470.yaml
```diff
@@ -0,0 +1,12 @@
+supportsClassB: true
+classBTimeout: 250
+pingSlotPeriod: 512
+pingSlotDataRateIndex: 2
+pingSlotFrequency: 0
```
johanstokking, line 5: This doesn't seem right

vendor/stmicroelectronics/nucleo-wl55jc-profile-as923.yaml
```diff
@@ -0,0 +1,12 @@
+supportsClassB: true
+classBTimeout: 250
+pingSlotPeriod: 512
```
johanstokking, line 3: Ping slot period is at most 128 seconds

vendor/stmicroelectronics/nucleo-wl55jc1.yaml
```diff
@@ -0,0 +1,74 @@
...
+firmwareVersions:
+  - version: '1.0.0'
+    numeric: 1
+    hardwareVersions:
+      - '1.0'
+    profiles:
+      EU863-870:
+        id: nucleo-wl55jc-profile-eu868
+        lorawanCertified: true
+        codec: nucleo-wl55jc-codec
```
johanstokking, line 15: This is a codec for a specific example, I suppose, right?

Would it be useful to name this firmware version then? This is what we did here as well: https://github.com/TheThingsNetwork/lorawan-devices/blob/9bdc8c5e96a70246d18ff03a898937d212a23046/vendor/the-things-products/the-things-uno.yaml#L6

johanstokking (commented):
Thanks, mostly questions!

vendor/stmicroelectronics/nucleo-wl55jc-profile-as923.yaml
pierrefraboulet-st: Thats true, the correct value defined in the LoRaWAN device is 16seconds.
Im aligned to the enum list included in schema.json. And this list defines all values as windows interval of 30ms.
https://github.com/TheThingsNetwork/lorawan-devices/blob/ba7ceb4fd89b385928e19bb9d676c98682a8b8f9/lib/draft/schema.json#L100-L103

But you're right, it's not very clear and it would be better to put in this enum the values from 2^0 to 2^7.

vendor/stmicroelectronics/nucleo-wl55jc1.yaml
pierrefraboulet-st: Thats right, this codec is aligned on LoRaWAN_End_Node projects: [STM32CubeWL LoRaWAN_End_Node](https://github.com/STMicroelectronics/STM32CubeWL/tree/main/Projects/NUCLEO-WL55JC/Applications/LoRaWAN/LoRaWAN_End_Node)

But I'm quite surprised, because with the exception of the the-things-uno.yaml file, all the examples currently offered on the master branch use the version field with a version string, and not to refer to a project name.
On the other hand, I agree that there is a missing field to reference the firmware to use (name, url, ...).

johanstokking: True, but for development kits a real version number is not super useful.

We don't enforce any string format on version, so I'm fine using a (versioned) project name.

vendor/stmicroelectronics/nucleo-wl55jc-profile-as923.yaml
johanstokking: You're right about this, we fixed this. Please rebase on `master`

vendor/stmicroelectronics/nucleo-wl55jc-profile-cn470.yaml
pierrefraboulet-st: Are you talking about the ping_slot field, related to the previous comment, or is it another element specific to this file?

johanstokking: I'm referring to the frequency `0`.

vendor/stmicroelectronics/nucleo-wl55jc2.yaml
```diff
@@ -4,7 +4,7 @@ hardwareVersions:
   - version: '1.0'
     numeric: 1
 firmwareVersions:
-  - version: '1.0.0'
+  - version: 'LoRaWAN_End_Node v1.0.0'
```
johanstokking, line 7: Thanks, this is clear this way

vendor/stmicroelectronics/nucleo-wl55jc-profile-cn470.yaml
pierrefraboulet-st: The beacon/ping slot frequency is not fixed for CN470, AU915 and US915. Here, the Regional Parameter definition for the Class B configuration of CN470:
`Example: the first beacon will be transmitted on 508.3Mhz, the second on 508.5MHz, the 9th beacon will be on 508.3Mhz again.`
Each 128seconds, the beacon and ping slot configuration use a new channel, defined by 2 equations:
Beacon : In Regional Parameters specification, chapters xxx Class B beacon 
`BeaconChannel = ...`
Ping Slot : In L2 specification (v1.0.3) chapter 16.2 Frequency-hopping beacon transmission
`Class B downlink channel = ...`

So, its not possible to set only one value in this configuration. If you have an interesting idea in order to provide the whole table of 8 frequencies corresponding to the 8 channels of Class B, I'm interested :)

johanstokking: Hmm, right, in these bands it doesn't make sense to specify this.

johanstokking (approved):

johanstokking:
Thanks!

PR #45, opened 2021-01-08
Add Laird RS1XX sensors

greg-leach:
Requested by: greg.leach@lairdconnect.com

#### Summary
Adds support for the Laird RS1XX range of sensors.


johanstokking (changes requested):
Thanks @greg-leach, looks good.

For codecs, we don't support modules and `require` logic for security and performance reasons. Please see https://www.thethingsindustries.com/docs/integrations/payload-formatters/javascript/

cc @benolayinka

johanstokking (commented):
Thanks @greg-leach 
The scripts are huge! Not saying it's a bad thing, but we might run into performance issues.

Is it feasible for you to add codec examples (see https://github.com/thethingsnetwork/lorawan-devices#payload-codecs) so that the input and output can be validataed automatically?

johanstokking:
I'm afraid we now get script execution timeouts... probably because the scripts are so huge.

Is it feasible to reduce the code size? Do you guys need assistance?

See error log here: https://github.com/TheThingsNetwork/lorawan-devices/pull/45/checks?check_run_id=1735483951

greg-leach:
I'm going to try running the validation script locally - we have a lot of different data, enumeration and bitfield types, hence the file sizes. I'll let you know how things go.

johanstokking:
> I'm going to try running the validation script locally - we have a lot of different data, enumeration and bitfield types, hence the file sizes. I'll let you know how things go.

See the README with instructions for that. With Node.js 14 and Go 1.15.x installed, it's just `make validate`.

johanstokking (approved):
Thanks!!

PR #63, opened 2021-01-13
More rounding in Oyster and Yabby Edge uplink decoders

jamesl-dm:
#### Summary
Added more rounding of battery voltages

#### Changes
- ...
- ...


johanstokking:
Thanks James!

PR #71, opened 2021-01-15
Add MCCI devices

terrillmoore:
#### Summary

Adds first set of MCCI devices.

#### Changes

Adds the MCCI vendor folder `vendor/mcci`, and three devices (Catena 4430, Catena 4450 and Model 4841).

#### Notes for Reviewers

I have more devices to push, but I figured I'd get a review on this. MCCI doesn't yet have a LoRa Alliance Vendor ID, so I've omitted that from our entries. The docs are not clear about whether this is permitted, and it's also not clear how to obtain one -- I sent a query to the LoRa Alliance.


johanstokking:
> I have more devices to push, but I figured I'd get a review on this. MCCI doesn't yet have a LoRa Alliance Vendor ID, so I've omitted that from our entries. The docs are not clear about whether this is permitted, and it's also not clear how to obtain one -- I sent a query to the LoRa Alliance.

This all looks perfect to me, thanks.

You don't need a `vendorID`, only when you want to generate standard QR codes and link from those to here, for onboarding. When you join the LoRa Alliance, you get a vendor ID from them.

vendor/mcci/catena4430-profile.yaml
```diff
@@ -0,0 +1,53 @@
...
+#rx1DataRateOffset: 0
+# RX2 data rate index
+#rx2DataRateIndex: 0
+# RX2 frequency (MHz)
+#rx2Frequency: 868.525
+# Factory preset frequencies (MHz)
+#factoryPresetFrequencies: [868.1, 868.3, 868.5, 867.1, 867.3, 867.5, 867.7, 867.9]
+
+# Maximum EIRP
+maxEIRP: 16
```
johanstokking, line 33: Are you using this for all your regions? Because all regions refer to this file.

terrillmoore: Well... we're using the same Murata module everywhere. The max power I've measured on our modules with the RWC5020 is pretty low, 16 to 18 dBm. And then EIRP depends on the (customer) antenna. Is this supposed to be a regulatory thing or a report of fact about our software/hardware combination. (I had a massive amount of data to merge, so rather than I went on, rather than asking.)

In fact, my team did a review this morning and found a few broken links, etc. So before merging, we might want to fix those. Based on what `maxEIRP` is supposed to represent, we can fix that at the same time.  (We'll have the same profile ID on each of these thing independent of region, so this would just be a matter of cloning files and adjusting.)

terrillmoore: converting to draft since we want to touch up a few links.

johanstokking: Thanks for your quick response.

> Is this supposed to be a regulatory thing or a report of fact about our software/hardware combination

It's the latter. I was mostly asking if you were sure; I'm fine merging like this if you fixed the links.

When you mark the PR as ready for review, I'll have a look again.

terrillmoore:
Sorry, commit log is wrong for 48ac9b5 -- should be "fix *link* in lora-alliance URL in comment". Will correct momentarily.

terrillmoore:
"this time for sure"

johanstokking (approved):
Perfect, thanks

PR #86, opened 2021-01-21
Add Multitech xDot

ymgupta:
#### Summary
Add `MultiTech` devices to the Device Repository.

Raising this PR on behalf of `Carol` from `MultiTech`.


johanstokking (approved):
I'm OK with this. The xDot alone is a module which is not really a stand-alone device though.

It would be helpful if there was a firmware version added with an example application, and that a codec is defined, for those who are getting started with the module to get started quickly. Anyway, I'll merge this in this state.

PR #97, opened 2021-01-25
Add milesight-iot devices

ymgupta:
#### Summary
Added `milesight-iot` devices 
  - em300-th
  - em500-co2
  - em500-udl


vendor/milesight-iot/em300-th.yaml
```diff
@@ -0,0 +1,76 @@
+name: EM300-TH
+description: Temperature&Humidity Sensor
```
benolayinka, line 2:
```suggestion
description: Temperature & Humidity Sensor
```

benolayinka (approved):

johanstokking (approved):
Thanks. Since there are no examples provided of the `em300-th` codec, the codec isn't being validated.

If you add examples, the validation will fail, because it uses the legacy `function Decoder()` signature, and not `function decodeUplink()`. This is already addressed in `am107.js`:

```js
function decodeUplink(input) {
  var res = Decoder(input.bytes, input.fPort);
  if (res.error) {
    return {
      errors: [res.error],
    };
  }
  return {
    data: res,
  };
}
```

I'm approving now, but just so you know, please make sure the codecs will work with the new signature otherwise adding examples will result in validation errors.

PR #85, opened 2021-01-19
Add SenseCAP

jancee:
#### Summary

Add SenseCAP

#### Changes

- ...
- ...

#### Notes for Reviewers

...

#### Release Notes

- ...

PR #107, opened 2021-01-28
Add SenseCAP to vendor index

benolayinka:
#### Summary

Add SenseCAP to vendor index. Missing from https://github.com/TheThingsNetwork/lorawan-devices/pull/85/files

### DIFF
--- a/vendor/index.yaml
+++ b/vendor/index.yaml
@@ -1624,3 +1624,6 @@ vendors:
     name: ThingsOfficer Gmbh
     website: https://www.boatofficer.com/
     logo: boatofficer_logo.png
+
+  - id: sensecap
+    name: SenseCAP


vendor/index.yaml
```diff
@@ -1624,3 +1624,6 @@ vendors:
     name: ThingsOfficer Gmbh
     website: https://www.boatofficer.com/
     logo: boatofficer_logo.png
+    
+  - id: sensecap
+    name: SenseCAP
```
benolayinka, line 1629:
```suggestion
    name: Seeed Technology Co.,Ltd.
```

neoaggelos (approved):
Make sure to fix CI errors before merging

johanstokking (changes requested):
We already have "Seeed Technology Co., Ltd" in the vendor list.

The vendor ID is not being used, only for references. So the end result is that there are two vendors called "Seeed Technology Co., Ltd" and that is confusing.

benolayinka:
> We already have "Seeed Technology Co., Ltd" in the vendor list.
> 
> The vendor ID is not being used, only for references. So the end result is that there are two vendors called "Seeed Technology Co., Ltd" and that is confusing.

They submitted the profile as sensecap, so I renamed them sensecap. They can fix it later if they need to.

johanstokking (approved):

PR #138, opened 2021-04-20
Add validation to check images can me be decoded by golang

mjamescompton:
#### Summary

When the hugo website is added to this repository it will need to decode the images to resize them. There are some images that where causing errors in Hugo, this adds validation to catch any future issues and fixes the current ones

#### Changes

- Add `tool/validate-images` go package to attempt to decode a png, jpeg or jpg
- Extend `bin/validate.js` to exercute `validate-images`
- Fix iotsens images by compressing them with https://tinypng.com/

#### Notes for Reviewers

...

#### Release Notes

- ...


bin/validate.js
```diff
@@ -227,6 +245,12 @@ vendors.vendors.forEach((v) => {
           });
         if (endDevice.photos.other) {
           endDevice.photos.other.forEach((p) => {
+            requireImageDecode(`${folder}/${p}`)
+              .then(() => console.log(`${key}: ${endDevice.photos.main} is valid`))
```
johanstokking, line 249: This isn't `photos.main`

tools/validate-image/main.go
```diff
@@ -0,0 +1,81 @@
...
+
+import (
+  "bytes"
+  "image/png"
+  "image/jpeg"
+  "io/ioutil"
+  "os"
+  "path/filepath"
+  "strings"
+	"log"
```
johanstokking, line 11: Run `gofmt`

tools/validate-image/main.go
```diff
@@ -0,0 +1,81 @@
+package main
```
johanstokking, line 1: Add Apache license

johanstokking (commented):
I would assume that `requireDimensions()` does the job already, by loading the image. Do you know why this is different?

tools/validate-image/main.go
```diff
@@ -0,0 +1,95 @@
...
+	if _, err := os.Stat(file); os.IsNotExist(err) {
+		log.Fatalf("Decode Image Failed: file %v does not exist", file)
+	}
+
+	ext := strings.ToLower(filepath.Ext(file))
+	if ext == ".png" || ext == ".jpg" || ext == ".jpeg" {
+		success := testPNG(file)
+		if !success {
+			success = testJPEG(file)
+		}
```
johanstokking, line 89: So we allow PNG format in a JPG extension? Why?

mjamescompton:
@johanstokking I'm not sure what in the image is causing this and I was unable to recreate in javascript, seem to be a go only problem where `png.Decode` fails with this specific .png. 

Digging in a little more, I see the the compression type for these images is JPEG, so it looks like they just have the wrong extension.

Screenshot 2021-04-21 at 10 28 06

So i'm assuming the `requireDimensions` does not rely on the extension to work what format the image is.

johanstokking:
It's not a Go only problem and we shouldn't get mismatching image file types and extensions. That's pushing the problem to clients, like web servers serving the wrong mime-type. That browsers are tolerant for this isn't a reason to allow it. `requireDimensions()` doesn't check the image file type anyway, only the dimensions.

Can't you use https://github.com/sindresorhus/image-type, after `requireDimensions()`?

PR #152, opened 2021-05-01
Add Tektelic agriculture sensor

### LINKED ISSUE #151: Include Tektelic Agriculture Sensor
#### Summary
At present, only the Kona indoor sensor is available for selection when provisioning a device.  It would be good to include the Agriculture Sensor as well

#### Why do we need this?
Increases the number of sensors available from Tektelic

Only the Kona generic sensor

An option to select the Agriculture Sensor

Add a new device profile

Yes.

proffalken:
#### Summary
Add the Tektelic Agriculture Sensor - closes #151 

#### Changes
Added a new profile for the Tektelic Agriculture Sensor along with associated codecs etc.


vendor/tektelic/t00059xx-codec.yaml
```diff
@@ -0,0 +1,2 @@
+uplinkDecoder:
+  fileName: decoder_agriculture_sensor.js
```
Jaime-Trinidad, line 2: Hi @proffalken can you provide an example?

proffalken: @Jaime-Trinidad - sure, that's the one bit that I'm missing from the docs.

Should the example contain the raw payload, example output, both, or something else completely?

Jaime-Trinidad: @proffalken yes payload and expected decode uplink data according to your file, you can follow the example-codec.yalm it would be great if we can have it. Thank you

proffalken: OK, I'll find some time to do that in the next few days.

Jaime-Trinidad:
Hi @proffalken any update?

Jaime-Trinidad:
> > Hi @proffalken any update?
> 
> Apologies, life is getting in the way of plans, I'll try and get this done in the next few days, it's working out the payload stuff that's throwing me right now :)

Thank you very much!

proffalken:
@Jaime-Trinidad - could do with your help on  this one.

My payload is `0104056E0202035C09650D570B6700AD0B68A5` and that results in the following output:

```
{
  "ambient_humidity": 82.5,
  "ambient_light": 3415,
  "ambient_temp": 17.3,
  "soil_gwc": 0.5,
  "soil_moisture_raw": 1390,
  "soil_temp": 17.029974645024424,
  "soil_temp_raw": 860
}
```

What I can't work out is how to convert that to the format at https://github.com/TheThingsNetwork/lorawan-devices/blob/master/vendor/example/windsensor-codec.yaml#L6 as if I start removing bits from the payload, the values change!

What am I doing wrong here?

johanstokking:
> What I can't work out is how to convert that to the format at [https://github.com/TheThingsNetwork/lorawan-devices/blob/master/vendor/example/windsensor-codec.yaml#L6](https://github.com/TheThingsNetwork/lorawan-devices/blob/master/vendor/example/windsensor-codec.yaml?rgh-link-date=2021-05-11T11%3A43%3A33Z#L6) as if I start removing bits from the payload, the values change!

This should go under `data:`, just like `direction` and `speed` in the example you referred to.

The input should be as a byte array; `[0x01, 0x04, 0x05, 0x6E, ...]`

The `soil_temp` looks like a problem for floating point. By the nature of floating point on different systems, let alone varieties in JavaScript runtimes, the value may differ. Consider using `toFixed(4)` to use 4 decimals, or whatever is significant in this case.

proffalken:
> The `soil_temp` looks like a problem for floating point. By the nature of floating point on different systems, let alone varieties in JavaScript runtimes, the value may differ. Consider using `toFixed(4)` to use 4 decimals, or whatever is significant in this case.

Gah, I'd forgotten how rubbish JS was at this kind of thing :(

Thanks for the advice, I'll sort it out when I get a moment :)

johanstokking (approved):

johanstokking:
@proffalken I set you on the wrong foot here; `toFixed()` returns a string. That's nice for display but doesn't really work well if you want floating point values. Rounding by decimals needs to go with the classic way; `Math.round(X * 1eY) / 1eY`.

So I take the blame and fixed it for you. This is good to go now.

proffalken:
> @proffalken I set you on the wrong foot here; `toFixed()` returns a string. That's nice for display but doesn't really work well if you want floating point values. Rounding by decimals needs to go with the classic way; `Math.round(X * 1eY) / 1eY`.
> 
> So I take the blame and fixed it for you. This is good to go now.

Fantastic, thanks @johanstokking!

PR #156, opened 2021-05-07
Assets bucket

michalborkowski96:
#### Summary
Adds TTUI assets bucket, per https://github.com/TheThingsIndustries/design/issues/15

This template also contains an IAM user to access the bucket, so that later we can have a github action that updates the bucket automatically.

#### Changes
Adds CloudFormation template for the S3 bucket. We don't want any loose resources on our production accounts.

### DIFF
--- a/deploy/1_assets_bucket.yaml
+++ b/deploy/1_assets_bucket.yaml
@@ -1,0 +1,23 @@
+AWSTemplateFormatVersion: '2010-09-09'
+Description: Assets S3 bucket for the device repository
+Outputs:
+  AssetsBucket:
+    Export:
+      Name:
+        Fn::Sub: device-repository-prod-AssetsBucket
+    Value:
+      Ref: AssetsBucket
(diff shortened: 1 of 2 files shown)


deploy/1_assets_bucket.yaml
```diff
@@ -0,0 +1,61 @@
...
+      Name:
+        Fn::Sub: ttui-assets-SyncUserAccessKeySecret
+    Value:
+      Fn::GetAtt:
+        - SyncUserAccessKey
+        - SecretAccessKey
+Resources:
+  AssetsBucket:
+    Properties:
+      BucketName: ttui-assets
```
johanstokking, line 35: This doesn't belong here.

Shouldn't this be in `design`?

Also please use Terraform for new stuff.

michalborkowski96: I am against this because everything else web-related uses CloudFormation, but okay

PR #157, opened 2021-05-07
Relax script execution timeout in CI

johanstokking:
#### Summary

Relax script execution timeout in CI

#### Changes

This increases the permitted time to run to 1 second. This is because the continuous integration here should only be a sanity check; the actual permitted execution time will depend on where these scripts are used and what the configuration is.

For instance, most The Things Stack deployments are running on a server, which are orders of magnitude faster than GitHub Action workers.

cc @greg-leach


johanstokking:
Also reverted disabling the `rs1xx` codec

johanstokking:
Okay so now CI fails with:

```
laird: rs1xx-ext-temp-rtd-sensor: EU863-870: payload codec rs1xx-ext-temp-rtd-sensor-codec invalid
./vendor/laird/rs1xx-ext-temp-rtd-sensor.js:decodeDownlink: output {"data":{"batteryType":"Lithium","ledBle":1928,"ledLora":1,"msgType":"rtdConfig","options":"None","sensorAggregate":0,"sensorReadPeriod":35,"tempAlarmLimitHigh":-61.99,"tempAlarmLimitLow":55.99,"tempAlarmsEnabled":"False"}} does not match {"data":{"msgType":"rtdConfig","options":"None","batteryType":"Lithium","sensorReadPeriod":35,"sensorAggregate":0,"tempAlarmsEnabled":"False","tempAlarmLimitLow":-200,"tempAlarmLimitHigh":450,"ledBle":1928,"ledLora":1}}
```

Which I'm not getting locally...

johanstokking:
Turns out this issue was on our side.

Issue #160, opened 2021-05-11
Allow access to the design repo


mjamescompton:
#### Summary

Here the issue we have with design repo, this repository does not have rights to the private repo during a release.

```
Run cd website/
yarn install v1.22.10
[1/4] Resolving packages...
[2/4] Fetching packages...
error Command failed.
Exit code: 128
Command: git
Arguments: ls-remote --tags --heads ssh://git@github.com/TheThingsIndustries/design.git
Directory: /home/runner/work/lorawan-devices/lorawan-devices/website
Output:
Warning: Permanently added the RSA host key for IP address '140.82.121.4' to the list of known hosts.
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.
```

#### Why do we need this?

Make including the design repository easier

#### What is already there? What do you see now?

I've manually included the design repository files.

#### What is missing? What do you want to see?

design repository should be included via package.json and github actions should be given the right to pull.

#### How do you propose to implement this?

I'm thinking the solution will be something like this

```
- name: Checkout design repository
      uses: actions/checkout@v2
      with:
        repository: TheThingsIndustries/design
        token: ${{ secrets.GitHub_PAT }} # `GitHub_PAT` is a secret that contains your PAT
```

@johanstokking is it correct to give this public repository action access to a private repository? 

#### Can you do this yourself and submit a Pull Request?

yes, might need tokens and bots made for me

johanstokking:
Why do we need to reference a private package if the assets are public already?

Can't we publish an npm package with the assets, and reference that here? Or do we need the sources here?

kschiffer:
> Why do we need to reference a private package if the assets are public already?

The issue is that TTUI is currently part of the design repo, which also includes assets that we cannot make public, such as the fonts. We could move TTUI into a separate repository but then again, in a way I like that the repo is private since it's not really thought to be used by the public and it being private gives us a bit more freedom to experiment. :shrug: 

> Can't we publish an npm package with the assets, and reference that here? Or do we need the sources here?

I think what we'll do is upload an `.tar.gz` of the bundled files to our new S3 bucket. It is possible to reference that in `package.json` as well just like a regular module. This way would give us the freedom to keep TTUI private for now but still have the bundled files public for each release.

PR #161, opened 2021-05-11
Use a fallback for design repository exported to dist

mjamescompton:
#### Summary

As I thought we have some permission issues with using the design repo, to unblock the action and have something we can use i've made an export of the design repo that can be use locally.

#### Changes

- Include export of design repository 
- Update import to use local design repository

#### Notes for Reviewers

I've written up the issue which should follow this here https://github.com/TheThingsNetwork/lorawan-devices/issues/160


johanstokking:
We now have the assets in a publicly accessible location, right?

PR #162, opened 2021-05-11
Allow profiles to be loaded from other vendors

### LINKED ISSUE #130: Vendor neutral name for DIY & small manufacturer devices
#### Summary

Require a vendor neutral name for use with community devices

#### Why do we need this?

So DIY devices that use the compliant open source stacks like the MCCI LMiC can be selected without people ending up down the wrong path with vendors that supply components - like MCCI, RAK, Arduino etc

I'd propose that small scale manufacturers can also piggy back off the generic entries - lest you end up with few hundred entries with the same base configuration. This can include people using the various modules from vendors such as Microchip, Murara & RAK as the model.

Nothing generic

Somethign like "Compliant Generic device based on ...." without it being an essay

At present the selection is entirely by vendor which makes sense but will be hard to signpost users as they won't really know what to select. So an additional line on the UI would be useful to tell them what to search for.

I'm working on the entries, just need a reasonable name that TTI are happy with.

What's wrong with "Generic"? You've got that!

I could do "DIY" for self-build and "Based on" for those using SiPs

Yes

johanstokking:
#### Summary

Closes #130 

cc @descartes @TD-er @terrillmoore 

#### Changes

- This allows profiles to be loaded from a particular vendor by specifying a `vendorID`. The idea is that module makers (`microchip`, `multitech`) as well as LoRaWAN end device stack vendors (`armmbed`, `semtech`, `mcci` maintaining LMiC?) can define profiles, like `microchip/rn2483-v1-0-1.yaml`
- This removes the `vendorProfileID: 42` that some profiles were using mistakenly, probably copy pasted from the example so I updated the example as well

#### Notes for Reviewers

@neoaggelos this needs to be implemented in the Device Repository fetcher. I don't think it's a big deal, but please do have a look and let me know if you have concerns.


neoaggelos (approved):
Implementation in The Things Stack should be straightforward, indeed, and can be done in time for 3.13.0.

Since we bundle the Device Repository with TTS releases, there should no be any breaking changes for existing versions.

One thing we would need to make sure is that we are extra careful with changes made to the generic profiles after people are already depending on them for their devices. For the most part, I think CI should be able to catch this.

johanstokking:
> One thing we would need to make sure is that we are extra careful with changes made to the generic profiles after people are already depending on them for their devices. For the most part, I think CI should be able to catch this.

Yes indeed. These generic profiles are static and one can't be changing the LoRaWAN version, potentially breaking many devices.

PR #173, opened 2021-05-17
format the files

cndingtek:
#### Using prettier to format the codes after make validate, make validate fmt.

using prettier format the codes.

#### Changes

- ...
- ...

#### Notes for Reviewers

...

#### Release Notes

- ...


cndingtek:
Regarding the last rejected pull request, we use make validate, make validate fmt to check again, there is not warning or error notification as below attached log. Then we try to use prettier to fix some typing format mismatch to see if it can meet the requirement.
[log.txt](https://github.com/TheThingsNetwork/lorawan-devices/files/6493676/log.txt)

vendor/dingtek/df702.yaml
```diff
@@ -0,0 +1,156 @@
...
+    numeric: 13
+
+# Firmware versions (at least one is mandatory)
+firmwareVersions:
+  - # Firmware version
+    version: '1.0'
+    numeric: 1
+    # Corresponding hardware versions (optional)
+    hardwareVersions:
+      - '12'
```
cBashTN, line 18: Shouldn't the corresponding hardwareVersions be 
``` 
hardwareVersions:   
      - '1.2'  
      - '1.3'  
``` 
?  

I'm not sure.

PR #216, opened 2021-06-07
Update t00059xx-agriculture-sensor.yaml

jammiemil:
#### Summary
Update to Tektelic Yaml to properly define Firmware/Hardware Versions 

#### Changes
Previous Version does not work in TTS as the Firmware and Hardware versions are not related to eachother


Jaime-Trinidad (approved):
changes look fine

PR #249, opened 2021-07-07
LTG92 Dragino decoder

Jaime-Trinidad:
#### Summary

Decoder now is updated as Dragino information.

#### Changes

- latitude and longitude decoder

#### Notes for Reviewers

...

#### Release Notes

- ...


vendor/dragino/lgt92.js
```diff
@@ -2,26 +2,13 @@ function decodeUplink(input) {
...
-{
-  data.longitude=(bytes[4]<<24 | bytes[5]<<16 | bytes[6]<<8 | bytes[7])/1000000;//gps longitude,units: °
-}
-else
-{
- data.longitude=0;//gps longitude,units: °
-}  
+
+data.latitude=(bytes[0]<<24 | bytes[1]<<16 | bytes[2]<<8 | bytes[3])/1000000;//gps latitude,units: °
+data.longitude=(bytes[4]<<24 | bytes[5]<<16 | bytes[6]<<8 | bytes[7])/1000000;//gps longitude,units: °
```
johanstokking, line 11: This doesn't seem to account for negative values

Jaime-Trinidad:
@johanstokking I fix it to consider negative values, it will send 0.

johanstokking:
But these values _can_ be negative

Jaime-Trinidad:
@johanstokking last version still get values if msb is 0, this can solve #331 what do you think?

johanstokking:
> @johanstokking last version still get values if msb is 0, this can solve #331 what do you think?

Which last version?

I agree we shouldn't use MSB as signaling whether the value is set.

Still we don't account for negative values here.

Issue #110, opened 2021-02-01
The Things Uno and Node missing settings for EU868


htdvisser:
#### Summary

The Things Uno is missing some settings that break routing when using ABP for EU868.

#### What is already there? What do you see now?

Device profile https://github.com/TheThingsNetwork/lorawan-devices/blob/master/vendor/the-things-products/the-things-uno-868.yaml that doesn't align with what the Arduino library does https://github.com/TheThingsNetwork/arduino-device-lib/blob/master/src/TheThingsNetwork.cpp#L701-L722

#### What is missing? What do you want to see? How do you propose to implement this?

I think this should be sufficient:

```yml
rx2DataRateIndex: 3
factoryPresetFrequencies: [867.1, 867.3, 867.5, 867.7, 867.9, 868.1, 868.3, 868.5]
```

but it needs testing.

jpmeijers:
Something we need to check is if The Things Uno (like many other older AVR devices) resets its frame counters to 0 when it starts up.

jpmeijers:
Yes, The Things Uno with the SendABP example from the Thing Network Arduino Library resets its frame counters when the board restarts. In the following screenshot I power cycled the board at 8:20:30.
![image](https://user-images.githubusercontent.com/948278/106560439-b3d6d000-652f-11eb-945a-37b47550d597.png)

The Things Uno is not the only device that behaves in this way. The most older devices (2015/2016) do not save the frame counters, and will start at 0 at startup or when the device restarts.

benolayinka:
@rvolosatovs @adriansmares @elsalahy if you have an Uno can you test this? Seems like it should be sufficient to enter these parameters manually in TTS.

Jaime-Trinidad:
I have been testing The Things Uno  using ABP and adding manually parameters, it doesn't present break routing, the only thing like this type of devices frame counter start in 0 everytime is reset.

elsalahy:
@Jaime-Trinidad did the testing requested by @benolayinka.
Please assign me back if you need my assistance with something.

Jaime-Trinidad:
Parameters were added manually, tests never present break routing. Using works fine:
rx2DataRateIndex: 3
factoryPresetFrequencies: [867.1, 867.3, 867.5, 867.7, 867.9, 868.1, 868.3, 868.5]

@htdvisser can we close this issue? do we need more test?

htdvisser:
[...]

PR #252, opened 2021-07-08
Add ABP configuration to TTI devices

Jaime-Trinidad:
#### Summary

Add ABP configuration to avoid break routing when using ABP for EU868 #110 

#### Changes

- ...
- ...

#### Notes for Reviewers

...

#### Release Notes

- ...


vendor/the-things-products/the-things-uno-abp.yaml
```diff
@@ -0,0 +1,29 @@
+name: The Things Uno ABP
```
htdvisser, line 1: Shouldn't this just be a different profile for the existing device?

htdvisser (commented):
I'm not the expert on these devices (otherwise I'd have done it myself), but this doesn't look quite right to me yet.

htdvisser:
@johanstokking since you contributed quite a lot to these Arduino SDKs I think it's better if you reviewed this.

johanstokking (changes requested):
Agreed with @htdvisser's points:

1. There is no ABP example for The Things Node
2. The Things Uno has an ABP example (`SendABP`: https://github.com/TheThingsNetwork/arduino-device-lib/blob/ea60c59a726d4147aad878a928da5c8baa7faa50/examples/SendABP/SendABP.ino), and that should simply be a firmware version with a profile. This should be under the `the-things-uno` end device
3. RX1 delay is 1 second

Jaime-Trinidad:
I fixed files:

1. I delete ABP on Generic Node.
2. Added a firmware version with a profile in The Things Uno file.
3. Fixed RX1 delay to 1 second.

vendor/the-things-products/the-things-uno.yaml
```diff
@@ -2,6 +2,10 @@ name: The Things Uno
 description: The Things Uno is based on the Arduino Leonardo with an added Microchip LoRaWAN® module. It is fully compatible with the Arduino IDE and existing shields.
 hardwareVersions:
   - version: '1.0'
+    numeric: 1
+  - version: '1.0-ABP'
+    numeric: 2
```
johanstokking, line 7: The hardware is the same

johanstokking (changes requested):

vendor/the-things-products/the-things-uno.yaml
Jaime-Trinidad: ok, now is fixed, same hardware version

johanstokking (approved):

PR #254, opened 2021-07-09
Elsys fixed files

Jaime-Trinidad:
#### Summary

Elsys devices files try to connect to example profile, this is fixed in 4 devices. https://www.thethingsnetwork.org/forum/t/register-end-device-keeps-fetching-forever/49592 

#### Changes

- commented vendorID line
- ers co2 lite, ers eye, ers lite and ers files fixed.

#### Notes for Reviewers

...

#### Release Notes

- ...


htdvisser (approved):

htdvisser:
Maybe it would also be good to update `validate.js` to catch this if it ever happens again.

cc: @johanstokking

PR #259, opened 2021-07-20
Update oyster.js

### LINKED ISSUE #233: Remove GPS coordinates when there is no fix
#### Summary

Remove GPS coordinates when there is no fix.

cc @jamesl-dm

#### What do you see now?

The Oyster and Yabby codecs may return position values while these may be outdated.

Since The Things Stack Application Server from 3.13.3 is automatically publishing location solved events if the decoded payload contains coordinates, this may lead to unwanted side effects.

#### What do you want to see instead?

I don't want to see these position values in the fields that are used for valid values. Instead, I would put them under `cached`, and return a warning.

Current codec: https://github.com/TheThingsNetwork/lorawan-devices/blob/9350955eefd9bd6468047b560ec540ce93575be8/vendor/digital-matter/oyster.js#L11-L24

Suggested change is to first check whether the fix failed. If that is the case, set the position values in a `cached` field and return a warning.

So the result would be:

```json
{
	"data": {
		"cached": {
			"latitudeDeg": 4.8684,
			"longitudeDeg": 52.34546,
			...
		},
		...
	},
	"warnings": ["fix failed"]
}
```

Can review

jpmeijers:
If fix failed, add location into a cache sub-object.

#### Summary

Fixes #233 

When fix failed, write latitude, longitude, speed, heading into a cache object, indicating that it is a cached location and not a real time location.

#### Changes

- Latitude, longitude, speed, heading into cache
- Simplify boolean checks

#### Notes for Reviewers

Example payload for fixFailed=true: `85A8C5EBD8763F0B0301BE`
Example payload for fixFailed=false: `94A3C5EB77743F0B0000BE`

#### Release Notes

- ...

### DIFF
--- a/vendor/digital-matter/oyster-codec.yaml
+++ b/vendor/digital-matter/oyster-codec.yaml
@@ -1,18 +1,34 @@
 uplinkDecoder:
   fileName: oyster.js
   examples:
-    - description: Position update
+    - description: Position update fix valid
       input:
         fPort: 1
         bytes: [0x00, 0x30, 0xed, 0xec, 0x00, 0x32, 0x24, 0x45, 0x00, 0x00, 0xde]
       output:
         data:
+          type: 'position'
           batV: 5.55
           fixFailed: false
-          headingDeg: 0
           inTrip: false
+          manDown: null
           latitudeDeg: -32
           longitudeDeg: 116
-          manDown: null
           speedKmph: 0
+          headingDeg: 0
+    - description: Position update fix failed
+      input:
+        fPort: 1
+        bytes: [0x85, 0xa8, 0xc5, 0xeb, 0xd8, 0x76, 0x3f, 0x0b, 0x03, 0x01, 0xbe]
+      output:
+        data:
           type: 'position'
+          batV: 4.75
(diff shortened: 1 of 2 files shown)


vendor/digital-matter/oyster-codec.yaml
```diff
@@ -1,18 +1,36 @@
 uplinkDecoder:
   fileName: oyster.js
   examples:
-    - description: Position update
+    - description: Position update fix valid
```
Jaime-Trinidad, line 4: @jpmeijers as warnings is in return it needs an empty string in output ` "warnings":[] `

jpmeijers: Ahh ok. Tricky to figure this out. Was expecting to only change the js decoder, but realised I need to update the codec too, which I'm not certain about.

Jaime-Trinidad: yes, I'm was pushing the change and it makes me fetch and I see your change, thank you!

Jaime-Trinidad (approved):

PR #267, opened 2021-07-27
Update tektelic devices

Jaime-Trinidad:
update tektelic devices

#### Summary

Support to update Tektelic devices.

#### Changes

- Remove old generation devices, yaml, codec, decoder files
- Update images

#### Notes for Reviewers

...

#### Release Notes

- ...


johanstokking (approved):
I don't see mistakes, but we're okay with keeping older generation devices in the device repository, because they may still be used.

Jaime-Trinidad:
> I don't see mistakes, but we're okay with keeping older generation devices in the device repository, because they may still be used.

You are right we don't discuss this point, I just have a quick call with them so I won't merge this. I will modify only to update device images and descriptions.

Issue #18, opened 2020-11-09
Command-line tools to create named device, profile and codec stubs


benolayinka:
#### Summary

It would be convenient to have a `make new vendorId device` command that adds the device name to the vendor index and generates template files that will fail `make fmt validate`, so lazy users know what they need to edit.

#### Why do we need this?

Slightly easier to add devices

#### Can you do this yourself and submit a Pull Request?

@johanstokking

johanstokking:
Indeed, having some basic CLI support would be useful.

johanstokking:
@Jaime-Trinidad do you feel comfortable picking this up?

Would be nice to have a simple Node based CLI (https://www.npmjs.com/package/node-cli) that generates stubs for vendor, end device, profiles, codec definitions etc based on templates (something like https://www.npmjs.com/package/text-template).

Jaime-Trinidad:
> @Jaime-Trinidad do you feel comfortable picking this up?
> 
> Would be nice to have a simple Node based CLI (https://www.npmjs.com/package/node-cli) that generates stubs for vendor, end device, profiles, codec definitions etc based on templates (something like https://www.npmjs.com/package/text-template).

@johanstokking Yes, I was thinking something like this will be usefull to device makers I work on it.

Jaime-Trinidad:
@benolayinka @johanstokking I have been working on this V1, for the moment it download device templates, I am working on v.2 so information can be written on CLI and generate each template https://www.npmjs.com/package/devrepo-templates

johanstokking:
Cool. Would be great if we can prompt a few things and fill them out.

Also why write the files under a `src` dir?

Jaime-Trinidad:
Now it download and fill the template, I just update the version 2.1.4 https://www.npmjs.com/package/devrepo-templates

johanstokking:
OK, so what is the plan with this? It's in your personal GitHub account now; we should either adopt it here, polish it up and make it part of the documentation flow, or keep it entirely separate.

Jaime-Trinidad:
I'm making some changes, so device makers can clone the Github and use it, it create the templates so they can paste them in their folder, I will polish it and make it part of documentation, also move it to TTN Github.

NicolasMrad:
any updates here?

NicolasMrad:
@Jaime-Trinidad what are the updates here?

johanstokking:
We're gonna do this differently.

PR #282, opened 2021-08-18
Template Generator

Jaime-Trinidad:
#### Summary

using `make file` command can generate and fill device repository templates. Folder `template-generator` contains a `README.md` file with an example and instructions. attends issue #18 

#### Changes

- Generate and fill templates

#### Notes for Reviewers

This is a draft, waiting for reviewers comments.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,3 +41,7 @@ fmt:
 	$(NPM) run format
 
 # vim: ft=make
+
+.PHONY: file
+file:
+	node ./template-generator/bin/create-project	
--- a/template-generator/README.md
+++ b/template-generator/README.md
@@ -1,0 +1,56 @@
+# Template Generator
+
+This is a Command-line tool to generate and fill Device Repository templates.
+
+## Documentation
+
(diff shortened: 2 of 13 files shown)


Makefile
```diff
@@ -41,3 +41,7 @@ fmt:
 	$(NPM) run format
 
 # vim: ft=make
+
+.PHONY: file
+file:
```
benolayinka, line 46: Without `npm install` this will fail on fresh clones of the folder.

I dont think you need the template generator to be a separate npm project, but @johanstokking will know better than me. If you consolidate it in the root folder, then `make deps` will already handle this with `npm install`. Otherwise, you need to add something in the makefile to cwd to your project and run `npm install`

benolayinka (commented):
I think a better flow for device creation is something like

- `make device`
- "Welcome to the device repository device template wizard. I'll help you create a template for adding a device. Select device vendor, or create a new vendor:
- Dropdown with all existing vendors, and option at top to create new vendor
- "Device name?"
- "Filename?" (or just create it by slugifying the device name)
- All your existing questions (hw, fw, certified)
- "Create regional profile?"
- "Create another regional profile?"
- "Create Payload Formatter codec template?"

PR #321, opened 2021-09-30
Fix ping slot frequency for SenseCap Lora E5

benolayinka:
https://forum.seeedstudio.com/t/example-for-adding-grove-lora-e5-to-ttn/259632/3

#### Summary

Bug in LoRa E5 profile


jancee:
To let the template work, this PR should be approved to support Class A.

Jaime-Trinidad:
@benolayinka class A and class B are on separated profiles, can you review it?

PR #323, opened 2021-10-06
add parametric devices

lmettler:
#### Summary

Added parametric devices with payload decoders.

#### Changes

- ...
- ...

#### Notes for Reviewers

...

#### Release Notes

- ...


lmettler:
Some additonal Information and a small bug in the decoder needs to be updated

PR #338, opened 2021-10-29
Add vendor ioThings and device ioTracker 3

Bertje:
#### Summary

Add vendor ioThings and device ioTracker 3

#### Changes

Initial addition

#### Notes for Reviewers

Replaces this PR: https://github.com/TheThingsNetwork/lorawan-devices/pull/332

#### Release Notes

N/A

PR #353, opened 2021-11-22
ioThings and ioTracker 3

Jaime-Trinidad:
#### Summary

Add ioThings vendor and ioTracker 3 from #338 

#### Changes

- ...
- ...

#### Notes for Reviewers

...

#### Release Notes

- ...


johanstokking (commented):
I don't see clear mistakes, but I do feel like there is a lot of redundant comments in these files. Those examples don't belong in the definitions. We have the documentation for that.

Also there are exceptions thrown when errors occur, while the nice solution is to return the errors in the `errors` array.

Bertje:
Thanks for your review @johanstokking, much appreciated.

Both remarks are understood and noted. Must both notes be fixed before a merge? Or would it be possible to merge now and have a cleaner/correcter version next time. We're eager to have our device available in the device repo, obviously, so please let us know what is the fastest, sufficiently clean route for us to take 😉

johanstokking:
@Bertje thanks for your comment. If you can update before tomorrow (Thursday) end of day (Amsterdam time) it will land in the same release of The Things Stack as when merging now. Otherwise it will be three weeks later.

Jaime-Trinidad:
Hi @Bertje can you update the files so we can merge them?

Bertje:
Yes: https://github.com/TheThingsNetwork/lorawan-devices/pull/388/

johanstokking:
Replaced by #388

PR #388, opened 2022-02-25
Add ioThings vendor and ioTracker 3 device

Bertje:
#### Summary
Cleaned/deleted up comments and updated uplink parser

#### Changes
- Cleaned/deleted up comments
- Updated uplink parser error handling and added support for man down features

#### Notes for Reviewers
See https://github.com/TheThingsNetwork/lorawan-devices/pull/353 for latest review status.


Bertje:
Thanks for the support Jona. We don't see that option, I think because we forked from an organisation account rather than from a personal account. We'll rebase and update on our end then. Or would you like to have temporary access to our fork in our organisation?

Bertje:
@johanstokking I rebased locally and updated the PR (checks passed now again as well).

johanstokking:
@Bertje thanks. The problem was that this pull request was targeting a feature branch here that wasn't up to date.

johanstokking (approved):

Issue #395, opened 2022-03-14
Enforce a standard format for decoded sensor data


pablojimpas:
#### Summary

The main explanation of this issue has already been described in a previous one: https://github.com/TheThingsNetwork/lorawan-devices/issues/237

However, this one is about making a detailed data model specification for the sensors in devices.

#### Why do we need this?

There's value in going a step further than just giving best practices, actually enforcing a strict format has benefits.

For example, suppose that someone wants to create an app that scans the QR code of a LoRaWAN device, automatically registers it in The Things Stack and starts displaying data from the device's various sensors. Right now, it's impossible to build such app in a manufacturer-agnostic way, completely decoupled from the device specifics.

Having a known data model/format will allow building integrations/apps using data coming from The Things Stack without caring about what device it's providing that data, the units used...just caring about the capabilities/sensors that it has.

#### What is already there? What do you see now?

Previous related issue: https://github.com/TheThingsNetwork/lorawan-devices/issues/237

#### What is missing? What do you want to see?

A strict data model specification, it's implementation in every current decoder and it's enforcement in new devices.

#### How do you propose to implement this?

  - In the spirit of simplicity I would enforce all units to be in SI, that's one less problem to care about when defining the data model.
  - Of course this enforced data model it's only for sensor data that the devices send, manufacturers would still be able to send all the additional data that they want in whatever format they wish.

#### Can you do this yourself and submit a Pull Request?

I can help providing requirements for the spec and maybe migrating current decoders to comply with the format.

johanstokking:
@pablojimpas thanks for your suggestion.

I think this can be very useful indeed. In my view, there would be a common JSON schema for decoded uplink and downlink data. Then, there would be another function that maps the output from the uplink decoder to this new schema.

I would like to keep that as a separate, optional step, after the uplink decoder. The decoded uplink is preserved in upstream messages in The Things Stack, and there will be a new field with normalized fields. The reason for this is that this standard JSON schema will never fully cover all device specific fields.

We already track [sensors](https://github.com/lorawan-schema/draft-devices/blob/master/schema.json#L885-L955) so what's really to be done is drafting a JSON schema with properties for each of these sensor types, that clearly define their data type, unit, description and validity (max/min, scale, patterns, enum values etc).

When we have that, we can define a new function signature maybe `normalizeUplink(data)` that simply takes the output from `decodeUplink()` and returns the normalized fields.

[...]

PR #400, opened 2022-03-28
WIP: Normalized payload

pablojimpas:
#### Summary

First steps for implementing #395 

#### Changes

  - Added `uplinkNormalizer` (similar to the previous `uplinkDecoder`)
  - Started work on the new `normalizedPayload` schema, for now just with common quantities and sensor.

#### Notes for Reviewers

  - This is just a work in progress draft, needs input and comments from all the interested parties.
  - I'm not very familiar with JSON schemas nor the lorawan stack in general, so be aware.
  - Some lines not relevant to this PR were reformatted by my editor, I did run `npm run format` but it didn't make a difference.

#### Release Notes

- Added the field `normalizedPayload` to standardize quantities and units, it is optionally created by the function `uplinkNormalizer` as an extra step in the decoding pipeline.

### DIFF
--- a/lib/draft/schema.json
+++ b/lib/draft/schema.json
@@ -15,7 +15,9 @@
     },
     "logo": {
       "allOf": [
-        { "$ref": "#/definitions/localFile" },
+        {
+          "$ref": "#/definitions/localFile"
+        },
         {
           "type": "string",
           "pattern": "\\.(png|svg)$"
@@ -24,7 +26,9 @@
     },
     "photo": {
       "allOf": [
-        { "$ref": "#/definitions/localFile" },
+        {
+          "$ref": "#/definitions/localFile"
+        },
         {
           "type": "string",
           "pattern": "\\.(png|jpg|jpeg)$"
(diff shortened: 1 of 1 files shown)


lib/draft/schema.json
```diff
@@ -343,6 +357,192 @@
             }
           ]
         },
+        "normalizedPayload": {
+          "allOf": [
+            {
+              "$ref": "#/definitions/endDevicePayloadCodec/definitions/scriptOutput"
+            },
```
johanstokking, line 364: This should be in `normalizePayloadOutput`

So there's:

- `normalizedPayload` with only the normalized payload properties

- `normalizePayloadOutput` which is the output of the normalize payload script (that has an `allOf` of script output as well as `normalizedPayload`)

- `uplinkNormalizer` under `endDevicePayloadCodec` that takes `decodeOutput` as input and `normalizePayloadOutput` as output

johanstokking (changes requested):
This PR should target https://github.com/lorawan-schema/draft-devices/blob/master/schema.json (the file in this repository is a git submodule or a subtree)

Please do not use your editor's default code styler for formatting. You can use `prettier` if it respects `.prettierrc` in the project root.

lib/draft/schema.json
pablojimpas: Okey! Thanks for your comments, I'll make changes accordingly.

Issue #405, opened 2022-03-30
Check that Vendor Profile ID for a vendor is unique 


KrishnaIyer:
#### Summary

Check that `VendorProfileID` for profiles of a vendor is unique. Refs; https://github.com/TheThingsNetwork/lorawan-stack/pull/5323

#### Why do we need this?

To make sure that we can query LoRaWAN device profiles with the (`VendorID`,`VendorProfileID`) tuple.

#### What is already there? What do you see now?

Currently, most device profiles don't have the `VendorProfileID` defined. For the few ones that are defined, the values are basically copy-pasted into multiple profiles. 

#### What is missing? What do you want to see?

If `VendorProfileID` is defined, it must be a unique value for each profile of a given vendor.

#### How do you propose to implement this?

I guess you'd need to ask users. @Jaime-Trinidad: Let's do this as an example for `The Things Products` and `The Things Industries` devices.

#### Can you do this yourself and submit a Pull Request?

Can test.

Jaime-Trinidad:
@KrishnaIyer I add a unique `VendorProfileId` in `The Things Product` and `The Things Industries` profile files as an example for this. 
I can contact makers to add numbers or I can start adding to the profile files.

Jaime-Trinidad:
@KrishnaIyer yes, the question `vendorProfileId` should be unique in each vendor profile or overall vendors. ex. `tti-profile-868.yaml` `vendorProfileId: 868` and `elsys-868.yaml` `vendorProfileId: 868` .
Here they can insert the same profileId different vendor. If they can't be repeated it will be difficult maybe it will be necessary to establish some rules about this, else if it doesn't matter because the vendor is different, only establish the rule that each profileId should be different.

KrishnaIyer:
> yes, the question vendorProfileId should be unique in each vendor profile or overall vendors. 

It should be unique within each vendor.

johanstokking:
I'll add validity checks in CI.

NicolasMrad:
Any updates here?

johanstokking:
We have two options here:

1. According to TR005, the profile ID can refer to a device profile and codec. So at least the combination of profile and codec must be unique. What we can do is declaring the numeric profile ID in the vendor's index file, referencing the region, profile and codec. This aligns with TR005, but has the downside that there are potentially many end devices using the same vendor profile ID, which makes onboarding challenging

[...]

PR #409, opened 2022-04-05
vendorProfileId

Jaime-Trinidad:
#### Summary

Add unique `vendorProfileId` to The Things Product and The Things Industries devices #405 

#### Changes

- Add a unique `vendorProfileId` in each profile file, without repeating the Id
- The Id is a random number

#### Notes for Reviewers

...

#### Release Notes

- ...


KrishnaIyer (commented):
@johanstokking: We are free to assign the `vendorProfileID` right? Should we follow some internal convention?
Just use linear numbers for each new profile?

johanstokking (commented):
Indeed we're free to assign them. They're 16 bit. This is already enforced by the JSON schema validator.

I would indeed make them increasing and not jump to 868.

KrishnaIyer:
> we can establish that 'vendorProfileId' should be filled linear instead of freely choice, what do you think?

 Ideally users can do what they want and we only check uniqueness but if that's too complicated to validate, then I guess it's ok to make it linear.

johanstokking:
We can make this part of validation in CI. 

Can you file an issue @Jaime-Trinidad ?

Jaime-Trinidad:
Yes I file an issue

PR #417, opened 2022-04-12
Add: add seeed lora e5 development boards

chenxixiya:
#### Summary

Add seeed lora e5 development boards

#### Changes

- ...
- ...

#### Notes for Reviewers

...

#### Release Notes

- ...


chenxixiya:
Hello, I have signed the CLA. But there is an additional account "jianfengyu" shown up, which is the same email as "chenxixiya". Will that be a problem? I couldn't find a way to delete it. What should I do about it? Thanks

Jaime-Trinidad:
> Hello, I have signed the CLA. But there is an additional account "jianfengyu" shown up, which is the same email as "chenxixiya". Will that be a problem? I couldn't find a way to delete it. What should I do about it? Thanks

yes it is strange let me check what can it be done

chenxixiya:
Maybe I can start a new one and re-push it, to see what it happens.

Jaime-Trinidad:
> Maybe I can start a new one and re-push it, to see what it happens.

Can you do that, to see if it is that.

PR #425, opened 2022-04-19
Fix decoder for rssi value

netvox-tech:
#### Summary

Fix decoder for rssi value

R718MA
R311D
R730MA


package-lock.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "lorawan-devices",
   "version": "1.0.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
Jaime-Trinidad, line 4: @netvox-tech please run `make deps` so this file doesn't get any modification

netvox-tech: OK, I will do that

johanstokking: You're probably using an older Node version. We're using Node 16 and npm 8.

netvox-tech: I have updated Node, npm, and go. It seems that the problem is fixed.
Thanks.

Jaime-Trinidad (approved):

PR #459, opened 2022-06-16
Fix BLE decoding

Bertje:
#### Summary
Fix in BLE payload decoding

...

#### Changes
- Fix in BLE payload decoding

#### Notes for Reviewers
...

#### Release Notes

- ...


johanstokking:
@Bertje I rebased your branch, can you please verify the diff is correct.

Also can you please comment why you update the LoRaWAN version of an existing device? Why not add a new firmware version with that new LoRaWAN version?

This may break existing devices in the field.

Bertje:
Thanks @johanstokking ; the diff looks good.
No breaking changes are introduced with the upgrade to the new LoRaWAN version. Thanks for checking.

PR #466, opened 2022-07-05
Updated Tektelic sensors

JBischoff1:
#### Summary

Updated the sensors for the TEKTELIC vendor

#### Changes

Updated the following sensors:
- Agriculture sensor -----------> CLOVER and KIWI
- Tundra sensor ---------------> TUNDRA
- Industrial GPS asset track ---> ORCA
- AC outlet gen2 ---------------> FLUX
- AC switch gen2 --------------> AURA
- Enterprise asset tracker ------> SPARROW
- Smart room base gen3 -------> COMFORT
- Smart room pir gen3 ---------> VIVID

Removed the following sensors:
- Industrial transceiver
- AC outlet gen1
- AC switch gen1
- Smart room base
- Smart room pir

Added the following sensors:
- eDoctor Device
- eBeat Device
- PELICAN EX
- PELICAN
- ROBIN
- FINCH

#### Notes for Reviewers

Note that the `make validate` command fails as the macVersion and regionalParametersVersion are causing problems even though they are included in the list of available versions in the README.

#### Release Notes

- ...


JBischoff1:
I was wondering if it is possible to add sensors to this device repository if they do not have decoder or encoder files?

JBischoff1:
@Jaime-Trinidad and @benolayinka when you have an opportunity, could you please review this pull request? Note that the validate script fails from the .regionalParametersVersion and .macVersion but the values I added are listed as valid in the README.

Jaime-Trinidad:
Hi @JBischoff1 some to fix first, rebase the PR, also do not delete devices, the reason is we do not know how many users have these ones, if we delete them it will cause a problem to those users.

JBischoff1:
@Jaime-Trinidad Thank you for your feedback! I have added the deleted devices back to the PR. Please let me know what other changes I should make.

johanstokking:
@JBischoff1 please see the errors returned by the Validate step: https://github.com/TheThingsNetwork/lorawan-devices/runs/7476366295?check_suite_focus=true

JBischoff1:
@johanstokking I am not sure why it is breaking as I mentioned above the validate script fails from the .regionalParametersVersion and .macVersion but the values I added are listed as valid in the README.

johanstokking:
> @johanstokking I am not sure why it is breaking as I mentioned above the validate script fails from the .regionalParametersVersion and .macVersion but the values I added are listed as valid in the README.

Yes they are listed, but the combination of LoRaWAN 1.0.3 with Regional Parameters for LoRaWAN 1.1 is not compatible, that is why. For LoRaWAN 1.0.3, the only permitted version is `RP001-1.0.3-RevA`.

JBischoff1:
@johanstokking Thank you. I fixed this for those sensors but the validate script still fails. The mac version and regional parameter version is now: 
macVersion: 1.1
regionalParametersVersion: RP001-1.1-RevB

johanstokking:
@JBischoff1 sorry this is a documentation thing on our end; `macVersion: 1.1` gets interpreted as a number, not a string.

I fixed this in the last commit here. Now there's another validation issue that I'm unable to solve; you're referencing `RP001-1.0.2-RevC` (but that should be `RevB` maybe?)

johanstokking:
Clarified our documentation in 9971c7b7e

JBischoff1:
@johanstokking Thank you for your help! I just made the fix and now the validation check passes. Please let me know if any other changes are needed.

johanstokking (approved):

PR #471, opened 2022-07-14
Added Senzemo (rebranding Sensedge to Senzemo)

sensedgeco:
#### Summary

Added Senzemo (rebranding Sensedge to Senzemo)

#### Changes

- Added Senzemo (rebranding Sensedge to Senzemo)
- Deleted Sensedge
- Added 3 devices

#### Notes for Reviewers

...

#### Release Notes

- ...


Jaime-Trinidad:
Hi @sensedgeco Thanks for adding the devices, I understand the rebranding but this way to do it should be not the best one, many users are under Sensedge selection on the LNS, if we simply erase and add this might bring problems with their connection, so let's create the new carpet and add the devices on it, also don't erase sensedge, and let's defuse this changes so users start changing their parameters on TTS

PR #485, opened 2022-07-26
Added vendor Senzemo

sensedgeco:
#### Summary

Added vendor Senzemo

#### Changes

- Added Senzemo SPU20
- Added Senzemo SMC30

#### Notes for Reviewers

...

#### Release Notes

- ...


vendor/index.yaml
```diff
@@ -1621,6 +1621,12 @@ vendors:
 
   - id: sensedge
     name: Sensedge
+    description: Measurements performed beautifully. Senzemo has +15 years of experience in IoT and experimental methods. We take great care in designing, testing and crafting our devices.
+    website: https://sensedge.co/
+    logo: sensedge_logo.png
+    social:
+      linkedin: https://www.linkedin.com/company/sensedge/
+      twitter: sensedgeco
```
johanstokking, line 1629: Are you using this vendor? Isn't this a duplicate of `senzemo`?

johanstokking (commented):
It would be great if you had some codec tests, see README with some pointers.

vendor/index.yaml
sensedgeco: Hi Johan,

The previous commit was rejected because we have removed the Sensedge vendor. The argument from the reviewer was that there are existing users that use old selection on the LNS and might bring problems with the connection. 

Regarding the codec, we are using the same one on multiple TTN applications and deployments already. 

Kr,
Luka

johanstokking: Yeah indeed it shouldn't be removed. I understand you changed your name. I think this is acceptable for now.

> Regarding the codec, we are using the same one on multiple TTN applications and deployments already.

Okay great.

johanstokking (approved):

PR #503, opened 2022-08-15
Fix Go source code formatting

adriansmares:
#### Summary

This PR formats the Go files using `gofumpt`. No review required.

### DIFF
--- a/website/tools/build/pkg/hugo/hugo.go
+++ b/website/tools/build/pkg/hugo/hugo.go
@@ -64,7 +64,7 @@ func copyImage(sourceFile string, destinationFile string, e *devicerepository.En
 			return err
 		}
 
-		err = ioutil.WriteFile(destinationFile, input, 0644)
+		err = ioutil.WriteFile(destinationFile, input, 0o644)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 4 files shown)


website/tools/build/pkg/hugo/hugo.go
```diff
@@ -103,8 +102,7 @@ func createFileFromTemplate(templatePath string, output string, tp *TemplateData
 
 // CreateContentSingleDevice generates the hugo content files in markdown
 func CreateContentSingleDevice(dir config.Dir, drs devicerepository.Store) error {
-
-	_ = os.Mkdir(dir.Hugo.Devices, 0700)
+	_ = os.Mkdir(dir.Hugo.Devices, 0o700)
```
mjamescompton, line 105: Out of interest @adriansmares what is the need for the lowercase `o` ?

adriansmares: It doesn't change anything semantically, it just makes things consistent with the rest of the notations that exist in Go (`0x` for hex, `0b` for binary). The formatter (`gofumpt`) does this automatically for consistency purposes. 

I recommend setting it in your VSCode settings as the formatter for Go in order to save time.

PR #510, opened 2022-08-22
Adding GWF RCM-LRW10 device

maebli:
adding gwf device rcm-lrw10


maebli:
I need to adjust the image as it is not official. I will make a new pull request and sign with the correct image

PR #515, opened 2022-08-24
Fixed fw-version format and removed optional info

rannolucST:
#### Summary

Shortened fw version field (max 32 char) and removed optional OperatingConditions info

#### Changes

- ...
- ...

#### Notes for Reviewers

...

#### Release Notes

- ...


johanstokking:
@rannolucST thanks for your submission. Please do not check in the executables in `bin`. I'll make sure they are ignored by `.gitignore`

rannolucST:
@johanstokking Ok, do we need another commit without the two executables or ok as it is for now?

johanstokking:
Looks very good otherwise!

Jaime-Trinidad (approved):

PR #524, opened 2022-08-31
Normalized schema additions

pablojimpas:
#### Summary

References #395 

#### Changes

- Introduce a new object for soil measurements.
- Add CO2 and light intensity properties to the air object.

#### Notes for Reviewers

- Some definitions are extracted, even though they are really simple. I've done this because I think it's a good practice to have quantities and their context decoupled as much as possible, that way quantities can be reused as the schema evolves. This is problematic in some cases, for example, the atmospheric pressure has a range that may not be suitable for industrial valves, in this case I think it makes more sense to maintain specific definitions inside each context.
- I've set the maximum of electrical conductivity based on the most conductive metal, silver. https://www.tibtech.com/conductivite. In practice, this is probably overkill.
- I haven't set a closed range for the pH scale based on this: https://pubs.acs.org/doi/pdf/10.1021/ed083p1465. In practice, maybe it makes more sense to set it to [0, 14] since most sensors in the edge aren't able to measure very high or very low values https://pubs.acs.org/doi/abs/10.1021/ac60120a014. I haven't extracted the pH because it is ridiculously simple with these assumptions. 
- I have only added a small batch of new measurements for now, let me know if you prefer larger PRs, I leave this as a draft until we agree on the right size.

#### Release Notes

- Added support for soil measurements, CO2 and light intensity in the normalized payload schema.


lib/payload.json
```diff
@@ -20,6 +20,23 @@
       "description": "Speed (m/s)",
       "minimum": 0
     },
+    "conductivity": {
+      "type": "number",
+      "description": "Electrical conductivity (dS/m)",
```
johanstokking, line 25: Why `dS/m` and not `S/m`?

lib/payload.json
```diff
@@ -34,6 +51,40 @@
...
+              "$ref": "#/definitions/temperature"
+            },
+            "EC": {
+              "description": "Soil electrical conductivity (dS/m)",
+              "$ref": "#/definitions/conductivity"
+            },
+            "pH": {
+              "description": "Soil pH level (open-ended scale)",
+              "type": "number"
+            },
```
johanstokking, line 72: I would say that `[0..14]` is a sensible practical range, indeed for the environment we're aiming at.

If we define validity (ranges, patterns) I think we should extract it to its own definition.

lib/payload.json
```diff
@@ -34,6 +51,40 @@
...
+          "properties": {
+            "moisture": {
+              "description": "Soil moisture (%)",
+              "$ref": "#/definitions/percentage"
+            },
+            "temperature": {
+              "description": "Soil temperature (°C)",
+              "$ref": "#/definitions/temperature"
+            },
+            "EC": {
```
johanstokking, line 65: Keys need to be lowerCamelCase, and if the first part is an abbreviation then it's lowercase; `ec`.

(otherwise uppercase, like `httpClient` vs `newHTTPClient`)

Same for `n`, `p` and `k` below.

lib/payload.json
```diff
@@ -50,6 +101,14 @@
               "description": "Atmospheric pressure (hPa)",
               "minimum": 900,
               "maximum": 1100
+            },
+            "CO2": {
```
johanstokking, line 105:
```suggestion
            "co2": {
```

johanstokking (commented):
Thanks! Few minor comments

I think it's useful to add a depth indicator to `soil` as well. It would be a value in centimeters.

lib/payload.json
pablojimpas: I choose `dS/m` because it's the most common nowadays for soil science:

- https://www.nrcs.usda.gov/Internet/FSE_DOCUMENTS/nrcs142p2_052803.pdf
- https://www.sdsoilhealthcoalition.org/technical-resources/chemical-properties/soil-electrical-conductivity/
- https://www.metergroup.com/en/meter-environment/education-guides/soil-electrical-conductivity-complete-guide-measurements
- https://www.fondriest.com/environmental-measurements/parameters/water-quality/conductivity-salinity-tds/

We could also use another more traditional unit like `μS/cm` or `mS/cm` which I think are more common in water measurements.

johanstokking: I see... so are we making the units domain specific (`dS/m` in `soil`), or do we go with the SI units everywhere (`S/m`)?

It also relates to the concentration; we could also use percentage there.

I tend towards going with domain specific units; so if soil moisture is expressed in `dS/m`, let's go with that. But if we do, then we shouldn't use a root definition for this.

lib/payload.json
```diff
@@ -20,6 +20,34 @@
       "description": "Speed (m/s)",
       "minimum": 0
     },
+    "depth": {
+      "type": "number",
+      "description": "Depth from the surface (cm)",
+      "minimum": 0
+    },
```
johanstokking, line 27: Every addition triggers new ideas.... I guess the first definitions here take time and then things fit in more naturally.

About depth; it's actually a position indicator. How about an object with longitude, latitude, altitude (above sea level) and accuracy? Those fields would come from a GPS receiver. That GPS receiver may not be in the same position as the sensor (if the GPS is above ground and the sensor is a meter under ground), so we may want to indicate a (relative) "level" (+/-) from that position. It's also useful for the wind sensor (which may be higher than the GPS receiver).

If we put a more generic level in the payload root (like now), developers may run into problems with combining measurements that are taken at different levels (i.e. wind sensor and a soil moisture sensor in one device). So would it be wise to assume that `depth` is the soil level (always positive, going from surface to ground) and so it should be in `soil`, to allow for an `elevation` in `wind`? Or would we have a generic `level` field in the root and if there are conflicts, require returning an array with different `level`s? I don't think that having both `elevation` and `depth` in the root is a good idea.

johanstokking: On this note; does the codec always know the depth in centimeters? Or do we have an alternative scalar value?

Our friends from @decentlab for instance have this DL-SMTP ([datasheet](https://cdn.decentlab.com/download/datasheets/Decentlab-DL-SMTP-datasheet.pdf)) device (which you may know) and it seems to use relative levels. 

BTW, about the soil moisture value that the sensor transmits:

>The soil moisture values are in the range of –5 to +120. The moisture sensors are calibrated to return
0.000 in air and 100.000 in water. Please note that the returned values ARE NOT soil moisture readings
in percents, but simply a scaled frequency unit (SFU). Please refer to AquaCheck documentation for the
conversion of SFU to volumetric water content.

Is this something that requires another field for soil moisture, or do you think that the conversion from this SFU to volumetric water content (%) can be done in the codec? Or would that need to be calibrated by the application?

lib/payload.json
pablojimpas: > I see... so are we making the units domain specific (`dS/m` in `soil`), or do we go with the SI units everywhere (`S/m`)?

It depends on how much we want to go down the rabbit hole and work to maintain the normalized schema:

- At one end, we can abstract everything using generic SI units for all properties, and then the application developer will do the conversion to whatever unit makes more sense for each context. However, this approach leaves much of the potential on the table for the new standardized payload.
- The other option is to try to be specific and carefully choose the units for each quantity-context combination, this will make the life easier for app developers in general, but edge cases could also appear.

Given that we are making the effort of creating this new format, I lean towards the second option, that way TTS will provide useful values for integrations out-of-the-box, without needing conversions in the application side, at least for the majority use cases.

lib/payload.json
pablojimpas: These are good points, I hadn't thought of them before.

Let's see, we already have an array of `measurements` in the normalized payload, would it be too much to ask for each device's normalization function to separate a payload into multiple `measurements` to avoid conflicts? For example:

This decoded payload:
```json
{
  "soilMoistureAtSurface": 11.2,
  "soilMoistureAtLevel6": 15.5,
  "windSpeed": 5.1
}
```

would become:
```json
[
  {
    "soil": {
      "depth": 0,
      "moisture": 11.2
    }
  },
  {
    "soil": {
      "depth": 60,
      "moisture": 15.5
    }
  },
  {
    "wind": {
      "elevation": 10,
      "speed": 5.1
    }
  }
]
```

That's leaving too much responsibility to the uplink normalizer function, but this is the first thing that came to my mind to solve this and support every use case possible.

Regarding the conversion units specific to the manufacturer, such us levels or scaled units from the calibration process, I think that those should be handled in the normalizer function as well (assuming that the info needed to make the conversion comes in the payload). The normalized format should only expose units that are "universally" relevant in the domains the data may be used, in the example above: SFU should be converted to volumetric water content (%). Otherwise, we will be leaking manufacturer specific details to applications, the same thing that we already have today with standard payloads.

johanstokking: > That's leaving too much responsibility to the uplink normalizer function, but this is the first thing that came to my mind to solve this and support every use case possible.

Yes this is also exactly what I think we should do.

---

> Regarding the conversion units specific to the manufacturer, such us levels or scaled units from the calibration process, I think that those should be handled in the normalizer function as well (assuming that the info needed to make the conversion comes in the payload)

I mean the situation where the end device isn't even aware of the depth. Let's say uncle George sticks a soil moisture sensor in his vinyard and only he knows he deep he put it in.

Just for clarity: I do still think we need a field for depth in centimeters. I just think we also need a field for a more generic "level" (number, starting from 0) that the application can map to a depth based on context.

In the long run, what else we should do in the device repository, is add (installation) state. Each device could declare what's relevant to know for an installed device (like depth), and the values would be part of the `input` to the codecs, so that the codecs can use those values as multipliers or references (if needed). In TTS Console, we could render a form to configure that state per device.

---

> The normalized format should only expose units that are "universally" relevant in the domains the data may be used, in the example above: SFU should be converted to volumetric water content (%). Otherwise, we will be leaking manufacturer specific details to applications, the same thing that we already have today with standard payloads.

Agreed

johanstokking (dismissed):

lib/payload.json
pablojimpas: > I mean the situation where the end device isn't even aware of the depth. Let's say uncle George sticks a soil moisture sensor in his vinyard and only he knows he deep he put it in.
> 
> Just for clarity: I do still think we need a field for depth in centimeters. I just think we also need a field for a more generic "level" (number, starting from 0) that the application can map to a depth based on context.

I am not sure if I get what you want here. If an end device doesn't support configuring the depth via downlink or other method, how are we supposed to know in the normalization phase what depth/level has to be assigned to that device? As far as I understand, the depth configuration has to either come in the payload if the device supports it or be handled by the integration completely. The normalization function does not have any kind of state, right? It works only with the input that it receives from the device.

johanstokking: It would be a logical value. Just like the DL-SMTP (that has six levels); no depth in centimeters but a number assigned to a sensor. You can't (and don't need to) send a downlink message with the depth in centimeters per logical level.

So I'm more looking after:

```json
[
  {
    "soil": {
      "level": 1,
      "moisture": 25
    }
  },
  {
    "soil": {
      "level": 2,
      "moisture": 32
    }
  }
]
```

In the future, if we have installation state:

```json
{
  "depths": [
    {
      "level": 1,
      "depth": 28
    },
    {
      "level": 2,
      "depth": 41
    }
  ]
}
```

And make that state available to the normalizer, then we can map it. For now, we can't but we need some way to convey non-depth-in-cm-levels I guess.

lib/payload.json
```diff
@@ -34,6 +56,46 @@
...
+              "$ref": "#/definitions/temperature"
+            },
+            "ec": {
+              "description": "Soil electrical conductivity (dS/m)",
+              "type": "number",
+              "minimum": 0,
+              "maximum": 621
+            },
+            "pH": {
+              "description": "Soil pH level (open-ended scale)",
```
johanstokking, line 81: It's not open ended anymore

lib/payload.json
johanstokking: I'll check with the Decentlab guys what their view is on this.

johanstokking: Got a reply from Decentlab. Bottomline is that soil measurements oftentimes need calibration and these parameters aren't (currently) available in the payload normalizer. For that we need per-device state. Let's not bother with SFU values in the normalized payload but wait until we can provide enough context to the normalizer for it to compute the VWC from SFU (i.e. the linear conversion parameters).

As for depth, forget about the relative depth: let's not do that indeed. The payload normalizer should have enough state to come up with the depth. I'm gonna add version identifiers (https://github.com/TheThingsNetwork/lorawan-stack/blob/feature/normalized-payload-warnings/api/identifiers.proto#L158-L162) to the normalizer's input, so that if there are different models with different depth configurations (like DL-SMTP), the normalizer can switch on the hardware version.

adriansmares: Regarding pushing the version identifiers to the codecs - do you think it will still be helpful when using the `vendorID/vendorProfileID/TBA` schematic ? because then it really becomes blurry what hardware is actually running the formatter / normalizer.

pablojimpas:
Sorry, I haven't looked at GitHub in the last few days.
I've fixed that pH description I left off last time.

Other than figuring out the depth and elevation issue, what else is missing for this PR? Is this an acceptable number of additions to the schema, or should we proceed in larger batches?

cBashTN:
Isn't the core idea of the normalizer to have standardized units? If yes, then why use `cm` instead of `m`?

lib/payload.json
```diff
@@ -20,6 +20,28 @@
       "description": "Speed (m/s)",
       "minimum": 0
     },
+    "depth": {
+      "type": "number",
+      "description": "Depth from the surface (cm)",
+      "minimum": 0
+    },
```
johanstokking, line 27: I find this acceptable now. We ditched the relative depth thing and we ditched making the version identifiers available to the normalized payload. If anything, we're gonna work with state in the future.

I do agree with @cBashTN that we should use common units as much as possible. I think there's reason to differentiate between % and ppm for concentration, but I would prefer using meters for depth (and elevation).

pablojimpas: I went with `cm` because it made more sense to me just for soil probes, kind of the same way that `hPa` are used for atmospheric pressure instead of `Pa`. But in this case it is a global definition that may be used in any other context, so meters are more appropriate then.

Should we use one definition for `depth` and another one for `elevation` and set the ranges accordingly, or just use one definition (e.g, `vertical`) to cover all cases allowing negative and positive values?

johanstokking: Okay, indeed we already use domain specific units (% and ppm). Let's keep things as domain specific as possible, so that we have soil depth in cm (and not generalize into `vertical`).

PR #527, opened 2022-09-02
Add device version to `normalizeUplink()` input

johanstokking:
#### Summary

References #395 
Preparations for https://github.com/TheThingsNetwork/lorawan-stack/issues/5759

cc @pablojimpas

#### Changes

- Add `elevation` to normalized `wind`
- Add `version` object to `normalizeUplink()` input, with information (and terminology) used in the Device Repository
- Extend `example` with how this can be used
- Allow for validating a single vendor

#### Notes for Reviewers

Review everything but last commit; that's just making sure everything is executed sequentially. It's slower, but it means less open file descriptors (because of concurrently running validation processes). Occassionally too many open file descriptors resulted in validation failure.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ deps.update:
 
 .PHONY: validate
 validate:
-	$(NPM) run validate
+	$(NPM) run validate -- --vendor="$(VENDOR)"
 
 .PHONY: fmt
 fmt:
--- a/README.md
+++ b/README.md
@@ -287,6 +287,8 @@ uplinkDecoder:
   # Examples (optional)
   examples:
     - description: 32 knots from the North
+      # Optional hardware versions for this codec example
+      hardwareVersions: ['1.0-2M']
       input:
         fPort: 1
         bytes: [0, 32]
--- a/README.md
+++ b/README.md
@@ -64,6 +64,12 @@ To validate data:
 $ make validate
 ```
 
+To validate only one vendor (much faster):
+
+```bash
+$ VENDOR=example make validate
+```
+
 [Visual Studio Code](https://code.visualstudio.com/) is a great editor for editing the Device Repository. You can validate your data automatically using the [YAML plugin](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml).
 
 The YAML plugin supports you with filling out the document. When hitting Ctrl + Space, all fields are shown. The Debug Console of Visual Studio provides feedback by highlighting the incorrect fields.
--- a/bin/validate.js
+++ b/bin/validate.js
@@ -192,38 +192,46 @@ function formatValidationErrors(errors) {
   return errors.map((e) => `${e.dataPath} ${e.message}`);
 }
 
-const vendors = yaml.load(fs.readFileSync('./vendor/index.yaml'));
+async function validate() {
+  const vendors = yaml.load(fs.readFileSync('./vendor/index.yaml'));
 
-if (!validateVendorsIndex(vendors)) {
-  console.error(`${options.vendor} index is invalid: ${formatValidationErrors(validateVendorsIndex.errors)}`);
-  process.exit(1);
-}
-console.log(`vendor index: valid`);
-
-const vendorProfiles = {};
-
-vendors.vendors.forEach((v) => {
-  if (options.vendor && v.id !== options.vendor) {
-    return;
+  if (!validateVendorsIndex(vendors)) {
+    console.error(`${options.vendor} index is invalid: ${formatValidationErrors(validateVendorsIndex.errors)}`);
+    process.exit(1);
   }
+  console.log(`vendor index: valid`);
 
-  const key = v.id;
-  const folder = `./vendor/${v.id}`;
-  if (v.logo) {
-    requireFile(`${folder}/${v.logo}`).catch((err) => {
-      console.error(`${key}: ${err}`);
-      process.exit(1);
-    });
-  }
+  const vendorProfiles = {};
 
-  const vendorIndexPath = `${folder}/index.yaml`;
-  fs.stat(vendorIndexPath, (err) => {
-    if (err) {
-      if (err.code !== 'ENOENT') {
-        console.error(`${key}: index file: ${err.code}`);
+  for (let v of vendors.vendors) {
+    if (options.vendor && v.id !== options.vendor) {
+      continue;
+    }
+
+    const key = v.id;
+    const folder = `./vendor/${v.id}`;
+    if (v.logo) {
+      requireFile(`${folder}/${v.logo}`).catch((err) => {
+        console.error(`${key}: ${err}`);
         process.exit(1);
-      }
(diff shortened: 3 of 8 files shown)


johanstokking:
@adriansmares https://github.com/TheThingsNetwork/lorawan-devices/pull/524#discussion_r960672057

> Regarding pushing the version identifiers to the codecs - do you think it will still be helpful when using the `vendorID/vendorProfileID/TBA` schematic ? because then it really becomes blurry what hardware is actually running the formatter / normalizer.

The codec and the profile are identified independently, also in the Alliance. The codecs should not have anything to do with the profile ID.

As for vendor ID, indeed that is an Alliance identifier that would be an alternative to the Device Repository's vendor ID (string). Actually, the `vendorID` here is probably just informative; as it stands now, the normalizer always belongs to the vendor (unlike profiles, which can be referenced from other vendors).

Note that the changes here are only applied to `normalizeUplink()`; we don't touch the decoder and encoders. Currently, the input and output structure of the decoder and encoders align with the LoRa Alliance TS013 specification. So we should avoid adding things there as that would diverge.

I think that in the payload normalization step, the actual hardware and region can be relevant. The firmware version and the payload encoding may be the same, but the hardware may use a different sensor or probe or have other physical attributes that affects how the payload is normalized. I don't think that profile ID is relevant.

adriansmares:
I'm thinking of the case in which the end device is onboarded via a QR code which contains only vendor ID and vendor profile ID. Unless I'm missing something, the payload formatter used by the Application Server won't have access to the end device version identifiers referenced here, because the AS cannot map back the `(vendor ID, vendor profile ID)` pair to the `(hardware version, firmware version, band ID)` pair that we're using in the device repository. 

How would payload normalization work here ? Do we just say that payload normalization is supported only when using the device repository format of the identifiers ?

johanstokking:
But in that case, how is the codec referenced in the first place?

As the identifiers are independent, there is already no way to go from profile ID to a codec, because we indeed miss hardware/firmware version. 

Unless the profile ID gets repurposed to a (unique) combination of profile and codec.

johanstokking:
If we repurpose that profile ID to also refer to the codec, then indeed we don't yet have the hardware and firmware version yet.

johanstokking:
Actually, TR005 already suggests that the VendorProfileID may refer to payload encoding as well:

Screen Shot 2022-09-03 at 17 12 31

Currently, the profile ID in the Device Repository identifies a device profile which can be used by many devices with different payload codecs. If we want to be compatible with TR005, we would have to key this profile ID to a device profile and codec ID. That would look like this:

```diff
diff --git a/vendor/example/windsensor.yaml b/vendor/example/windsensor.yaml
index b59a884c3..65f16a246 100644
--- a/vendor/example/windsensor.yaml
+++ b/vendor/example/windsensor.yaml
@@ -39,10 +39,12 @@ firmwareVersions:
         id: windsensor-profile
         lorawanCertified: true
         codec: windsensor-codec
+        vendorProfileID: 42
       US902-928:
         id: windsensor-profile
         lorawanCertified: true
         codec: windsensor-codec
+        vendorProfileID: 43
 
   - # You can add more firmware versions and use different profiles per version
     version: '2.0'
@@ -54,13 +56,16 @@ firmwareVersions:
         id: windsensor-profile
         lorawanCertified: true
         codec: windsensor-codec
+        vendorProfileID: 42
       US902-928:
         id: windsensor-profile
         lorawanCertified: true
         codec: windsensor-codec
+        vendorProfileID: 43
       AS923:
         id: windsensor-profile
         codec: windsensor-codec
+        vendorProfileID: 44
 
 # Sensors that this device features (optional)
 # Valid values are:
```

That is still too limiting for the Device Repository: multiple devices, and multiple firmware versions of a single device (that can run on multiple hardware versions), can use the same combination of profile and codec. So we still cannot go from `vendorProfileID` to our primary keys here: vendor ID, end device ID, firmware version, hardware version and region.

I think we should add this `vendorProfileID` field and drop `vendorProfileID` from the device profile. That way, we are at least compatible with TR005.

```diff
diff --git a/schema.json b/schema.json
index 778a73e87..494dbe796 100644
--- a/schema.json
+++ b/schema.json
@@ -620,12 +620,6 @@
         }
       ],
       "properties": {
-        "vendorProfileID": {
-          "type": "integer",
-          "description": "VendorProfileID managed by the vendor, as defined in TR005",
-          "minimum": 0,
-          "maximum": 65535
-        },
         "supportsClassB": {
           "type": "boolean",
           "description": "End device supports class B"
```

Our validation can ensure that these vendor profile IDs are the same for referenced profiles and codecs. This allows us to at least load the profile and codec when we have a QR code without ambiguities. What we still miss to identifying devices: we miss the hardware and firmware version.

However, this pull request now becomes problematic. If we _can_ load the codec (including the normalizer), but we don't have the hardware and firmware version and the normalizer requires these versions, the normalizer may return errors. That is not a good developer UX.

So what do we do?

1. Do not accept the changes here. Do not make `normalizeUplink()` take identifiers may not be known if the device has been onboarded with a QR code (and the owner didn't select the right device). This implies that normalizers that do need device specific information (unlike the decoder) cannot obtain this information. If the device maker has this foresight, they must declare multiple firmware versions for each hardware version and assign a new `vendorProfileID` to each unique hardware and firmware version (but is not advised in TR005)

2. Do accept the changes here. Let The Things Stack only run `normalizeUplink()` when the hardware and firmware versions are known. However, most normalizers probably don't require these versions to be present, so this still is not really nice, but I think it's acceptable. By the way, even if the owner didn't select the hardware and firmware version, it may still be discovered with [TS006 Firmware Management](https://resources.lora-alliance.org/document/ts006-1-0-0-firmware-management-protocol):
 
Screen Shot 2022-09-03 at 17 44 32

(Maybe we should also add TS006 support in the Device Repository, so our AS knows that it can discover these versions? This is nice especially with FUOTA, and works with TR005 for onboading and TS006 for version discovery and then we have all identifiers)

@adriansmares looking forward to hearing your thoughts. Let's make sure we have a proper plan before we merge this at least.

johanstokking:
Updated the above with progressive insight

adriansmares:
I think we should not add the device version as an input.  As things stand right now, it will only widen the indexing gap between our repository and the specification. TR005 encourages profile and codec reuse, and with reuse this functionality is useless. I somehow expect that if by magic normalizers become a specification, they will reuse the vendor profile ID too.

The customer UX that we will have with these QR codes in general is not ideal, but that battle has been lost on TR005 publication. 

In its current form, I believe that end devices that would like to have hardware / software model specific information available in the formatter or normalizer should just have an extra device profile ID / dedicated formatter. If we have to reuse the same vendor profile ID between two sensors that have different physical properties, it's not like we can do it in a stateless manner.

---

I like the refactoring that maps the vendor profile ID to a profile + codec - I think we should do that.

johanstokking:
OK, let's forget about this for the time being then!

PR #542, opened 2022-10-11
Delete duplicated

Jaime-Trinidad:
Thermokon was duplicated in the root

#### Summary

...

#### Changes

- ...
- ...

#### Notes for Reviewers

...

#### Release Notes

- ...


Jaime-Trinidad:
Hi @ysmilda I delete the duplicated can you review it so I can merge it

ysmilda (changes requested):
When the files were added to the root they were also updated, those changes are now lost. Can you copy the files into the `vendors` folder instead of just deleting them?

Jaime-Trinidad:
> When the files were added to the root they were also updated, those changes are now lost. Can you copy the files into the `vendors` folder instead of just deleting them?

Yeah, they are already there I verify before deleting them from the root, and are the same, that's why I didn't copy them to not duplicate

Jaime-Trinidad:
@ysmilda review done including some fixes

ysmilda (approved):
LGTM!

PR #561, opened 2022-11-26
Update esp32-paxcounter-codec

cyberman54:
Payload API for paxcounter devices changed a litte, a new value sleepcycle is delivered. This update catches up with it.


cyberman54:
cancelled, will come up with a new one.

Issue #421, opened 2022-04-13
Tolerance of null payloads


eggfriedrice:
#### Summary
We have a customer using a Browan TBHH. The decoder sometimes returns garbage numbers, and it appears that this happens because on some uplinks the `frm_payload` is absent. The decoder then miscalculates temperatures etc.

...

#### What do you see now?
Bad payload decoded with no `frm_payload`:

    "received_at": "2022-04-08T16:14:39.574932888Z",
    "uplink_message": {
      "session_key_id": "AX+P/WzBNi9gzOILCFnKcQ==",
      "f_port": 103,
      "f_cnt": 964,
      "decoded_payload": {
        "battery": 2.5,
        "humidity": 0,
        "status": 0,
        "temperature": -32
      }

 

Good payload decoded fine (note presence of `frm_payload`:

    "received_at": "2022-04-08T16:14:33.901128424Z",
    "uplink_message": {
      "session_key_id": "AX+P/WzBNi9gzOILCFnKcQ==",
      "f_port": 103,
      "f_cnt": 963,
      "frm_payload": "CAssVP////8=",
      "decoded_payload": {
        "battery": 3.6,
        "humidity": 84,
        "status": 1,
        "temperature": 12
 

#### What do you want to see instead?
The question really is, should the stack try to decode null payloads, or should the decoder functions be tolerant of this?

#### How do you propose to implement this?
Should the TBHH decoder be fixed to just return nothing if nothing is passed in?

#### Can you do this yourself and submit a Pull Request?

...

Jaime-Trinidad:
Hi @eggfriedrice how often is presented this issue, I think is not a null payload, also I see that is sending data every 5 seconds? I make some tests with `null` and `00` payloads:

``` "received_at": "2022-04-13T16:18:25.536691641Z",
    "uplink_message": {
      "f_port": 103,
      "frm_payload": "AA==",
      "decoded_payload": {
       "battery": 2.5,
        "humidity": 0,
        "status": 0,
        "temperature": -32
      },

eggfriedrice:
Hi @Jaime-Trinidad 
This is from a customer, so I can't verify much, but note in the first snippet I posted that there is no `frm_payload` field at all.

johanstokking:
@Jaime-Trinidad it looks like the decoder returns "base" values when the payload is empty. The decoder should return an empty object with the payload is empty, that's all.

NicolasMrad:
[...]

PR #609, opened 2023-04-24
Decoder files to return empty objects

Jaime-Trinidad:
#### Summary

Decoder files were fixed to return empty object #421 

#### Changes

- Check if bytes are present if not, return an empty object
- ...

#### Notes for Reviewers

...

#### Release Notes

- ...


vendor/browan/tbam100.js
```diff
@@ -1,6 +1,11 @@
 function decodeUplink(input) {
     var bytes = input.bytes;
 
+     //Check if temperature Board byte is present
```
adriansmares, line 4: I think you should add these checks inside the the switch case instead, before the return. Then you can keep the `unknown FPort` error.

KrishnaIyer:
@Jaime-Trinidad: Can you fix the review comments and complete the PR? If it's a lot of effort, I can find someone to take this over.

adriansmares (approved):

PR #614, opened 2023-05-15
Update Dragino Decoder

dragino:
1. Fix previous decoder bugs
2. Add Decoders for new models.


Jaime-Trinidad (approved):

Jaime-Trinidad:
@dragino please change the name of your files, they should be lower case, also check the files where in the codec name, that is the reason validation is failing

vendor/dragino/lse01-121-codec.yaml
```diff
@@ -0,0 +1,17 @@
+uplinkDecoder:
+  fileName: lse121.js
```
Jaime-Trinidad, line 2: update for the right name of .js file `lse01-121.js`

Jaime-Trinidad:
Hi @dragino please fix the file and run validation again. Thanks

PR #631, opened 2023-05-31
Fix of Nwave G4 decoders

nabrosimoff:


nabrosimoff:
Hi @Jaime-Trinidad . I've updated the branch

Issue #188, opened 2021-05-25
Add device repository homepage


mjamescompton:
#### Summary

Add a page that has an overview of most popular devices, vendors and tags. Featured vendors and an explanation of what can be found in the device repository. 

#### Why do we need this?

Making it easier to find devices you are looking for and improving SEO

#### What is already there? What do you see now?

The root page is identical to `/devices` which is redirected to the root. The root is just a list of all devices 

#### What is missing? What do you want to see?

The page should include, repository description, search, featured devices, featured tags, featured vendors, Link to all devices.

#### How do you propose to implement this?

Design by @pierrephz, @mjamescompton will build

#### Can you do this yourself and submit a Pull Request?

yes

KrishnaIyer:
This is covered by the partner portal: https://partners.thethingsindustries.com/devices

Issue #185, opened 2021-05-25
Add tag descriptions


mjamescompton:
#### Summary

Generate and include tag description

#### Why do we need this?

To improve SEO the tag landing pages need to have a description explaining them.

#### What is already there? What do you see now?

All Sensors, IPCode, Frequency plans have a tag page generated, if they are defined by a device.

#### What is missing? What do you want to see?

- Create content for all tags. @anastasiagramm 
- Add additional field as tags. @mjamescompton 
- Include tag description, either within the schema or this can be Hugo data @johanstokking what do you recommend?
- Display field on tag pages. @mjamescompton 

#### How do you propose to implement this?

@anastasiagramm  Could you work on the content for the tags, I have generated a google sheet 

https://docs.google.com/spreadsheets/d/1OC-bcU5a8ydJBKBt0QELO-MqzgA7nhoEoGSTFMZ4hvM/edit?usp=sharing

Can you fill in the descriptions here, there is a separate table for Sensors, IPCode, Frequency plans. I have included all the allowed fields from device schema 

https://github.com/TheThingsNetwork/lorawan-devices/blob/master/lib/draft/schema.json

so they should all be there and we can add new tables if we include any new fields. A small warning I will not read the descriptions so please make sure they are correct in this doc.

#### Can you do this yourself and submit a Pull Request?

yes

mjamescompton:
@anastasiagramm I have also added a field at the bottom of each page for a general description of the tag category Sensors, IPCode, Frequency plans

mjamescompton:
I've added LoRaWan version and Regional Parameters and a few others which are less important, if you can start with the ones above first and LoRaWan version and Regional Parameters, ignore the others for now.

johanstokking:
> * Include tag description, either within the schema or this can be Hugo data @johanstokking what do you recommend?

We can't really put it in the schema because JSON schema doesn't support descriptions for enum values.

I think that a data file would be helpful. I'm okay with checking that in at https://github.com/lorawan-schema/draft-devices. It can be validated against the schema.

```json
{
  "sensors": {
    "temperature": {
      "description": "Temperature"
    },
    ...
  },
  "ipCodes": {
    "IP68": ...
  },
  ...
}
```

anastasiagramm:
@mjamescompton I've filled in the descriptions that you mentioned. please let me know when you need the other ones

mjamescompton:
@laurensslats Can you review the descriptions

[...]

PR #690, opened 2023-11-01
Add tag dropdowns

LDannijs:
#### Summary

Add a dropdown menu that lists certain types of tags to filter through the devices.

Screenshot 2023-11-01 at 13 55 23
Screenshot 2023-11-01 at 13 55 36

#### Changes

- add rudimentary code for tag dropdown
- added a bit of styling (just for testing mainly)

#### Note
This is basically the same PR as #680 but i accidentally deleted the fork. Whoops. Please refer to that PR for the discussions around it.


pierrephz:
https://github.com/TheThingsNetwork/lorawan-devices/assets/26456318/1af8975d-90ef-4c01-98a2-a61fa551f1b8

here is a proposal for an uplifting of the Device Repository home page
Open for feedback

cc @KrishnaIyer

<details>

<summary>Expend to view the mockup</summary>

desktop-device-repo

</details>

KrishnaIyer:
@pierrephz: Where does the `Submit Your Device` button lead to?

mjamescompton:
These filter need to work for everything page type not just the homepage. 

For now I think we are good having a page that contains every single device in the repo but at some point this page will become very slow. This may in happen when we start to parse the json and render everything when we are not using our Shinny new M1's. At this point it might time to add a more curated homepage https://github.com/TheThingsNetwork/lorawan-devices/issues/188 and then spilt the devices up in to categories.

Our second most valuable page other then the homepage is https://www.thethingsnetwork.org/device-repository/tags/temperature/   (we can still get this page a little higher if we complete this task https://github.com/TheThingsNetwork/lorawan-devices/issues/185 @KrishnaIyer )

This tag page will also need these filters as there are already to many devices on this page.

For this work the filter need to populated based on the content on the page and the page need to contain all devices already in the html.

I think the best way to do this is

With a JSON object containing all the device data on a page (homepage, tag, or device maker) on the page. We would still have all the devices on the page in html when it loads, this json object will also be there ready for js to consume. When the filter dropdown is press we could rebuild the devices list from this json

The json object will be generated by hugo, I believe as we are already requesting the tag we will already have an object with the devices from that tag and it might even already have the attributes. We are already generating something similar to this json object on the single device page `website/layouts/devices/single.json.json` which is imported into the page 

```
<script>
  var config = {
    device: JSON.parse("{{ .Params | jsonify }}")
  }
</script>
```

We need to write the custom js to filter the elements.

PR #691, opened 2023-11-08
add minol-zenner vendor and edc module definitions

ant2alex:
#### Summary
this adds the minola-zenner vendor folder and a first version of the edc communications module with a fitting decoder.

#### Changes
- minola-zenner vendor added
- edc communication module profile and decoder added

#### Notes for Reviewers
- make validate ran only when removing "vendor/example" folder temporarily due to a failure in the image checking utility

### DIFF
--- a/vendor/minol-zenner/edc-communication-module-profile.yaml
+++ b/vendor/minol-zenner/edc-communication-module-profile.yaml
@@ -1,7 +1,7 @@
 # Vendor profile ID, can be freely issued by the vendor
 # This vendor profile ID is also used on the QR code for LoRaWAN devices, see
 # https://lora-alliance.org/sites/default/files/2020-10/LoRa_Alliance_Vendor_ID_for_QR_Code.pdf
-vendorProfileID: 474
+vendorProfileID: 1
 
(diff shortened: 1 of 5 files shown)


vendor/minol-zenner/edc-communication-module-profile.yaml
```diff
@@ -0,0 +1,21 @@
+# Vendor profile ID, can be freely issued by the vendor
+# This vendor profile ID is also used on the QR code for LoRaWAN devices, see
+# https://lora-alliance.org/sites/default/files/2020-10/LoRa_Alliance_Vendor_ID_for_QR_Code.pdf
+vendorProfileID: 474
```
LDannijs, line 4: I want to quickly notify you on changing this. The vendorProfileID is different from the VendorID. This vendorProfileID should be an incremental value, used for identifying devices within a vendor. So change it to something along the lines of 1, 101, etc.

I shall have that text changed to better reflect what it's supposed to be.

ant2alex: Thanks for the clarification. I changed the vendorProfileId to 1

LDannijs (commented):
The files look good to me, but I do not have "approval" status, only @Jaime-Trinidad can do that. I will let him know.

Jaime-Trinidad (approved):

PR #693, opened 2023-11-09
Add a setup script

LDannijs:
#### Summary

At the moment, users are required to create a bunch of files containing boilerplate code that they then need to edit whenever they want to add a vendor or new devices. This PR add 2 scripts that create all of the files by just entering a few questions. I'd like to receive feedback on what other changes could be made to improve it.

This is an in between feature, idea is that eventually a type of form will be used for the repo, but for now this should at least be helpful.

(recreated this PR cause i had some organizing of my fork)

#### Changes

- add 2 scripts. one for vendor + devices and one for just devices.

Preview of what the scripts look like.
image

image


Jaime-Trinidad (approved):

Jaime-Trinidad:
@johanstokking can you take a look to @LDannijs work here

setup
```diff
@@ -0,0 +1,317 @@
+#!/bin/bash
```
johanstokking, line 1: Extension should be `.sh`

setup
```diff
@@ -0,0 +1,317 @@
...
+
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+BLUE='\033[0;94m' 
+YELLOW='\033[0;33m'
+NC='\033[0m'
+
+echo -e "${BLUE}Hello! The Device Repository is where we store lora devices. To add a new vendor or devices, some files need to be created. To help you along we've created this setup tool that will create a few template files with the basic information needed. Please fill in the upcoming questions.\n${NC}"
```
johanstokking, line 9:
```suggestion
echo -e "${BLUE}Hello! The Device Repository is where we store characteristics and capabilities LoRaWAN devices. To add a new vendor or devices, some files need to be created. To help you along we've created this setup tool that will create a few template files with the basic information needed. Please fill in the upcoming questions.\n${NC}"
```

setup
```diff
@@ -0,0 +1,317 @@
...
+    echo -e "${BLUE}What is the LoRa Alliance issued Vendor ID? (${RED}required${BLUE}):${NC}"
+    read vendorid
+    if [[ $vendorid =~ ^[0-9]+$ ]] && [[ $vendorid -ge 0 && $vendorid -le 1000 ]]; then
+        if grep -q -w "vendorID: $vendorid" ./vendor/index.yaml; then
+            echo -e "${RED}This Vendor ID is already in use.\n${NC}"
+        else
+            break
+        fi
+    else
+        echo -e "${RED}Invalid input. Please enter a number between 0 and 1000.\n${NC}"
```
johanstokking, line 43: If they're not a member they can skip this

johanstokking (commented):
Nice.

To avoid duplication can you call the setup devices script from the setup script? You can pass the vendor ID as argument so you can skip that prompt.

johanstokking (commented):
Thanks.

Few more things:
1. Add line breaks before 80 characters (that is the max line length)
2. Respect `NO_COLOR` environment variable. If set, don't print colors. Not all terminals are able to show them and so it becomes an unreadable mess. You can keep the assignments but make sure they're empty when `NO_COLOR` is set

LDannijs:
> 1. Add line breaks before 80 characters (that is the max line length)

Would I need to do this for only the echo commands or the entire code?

LDannijs:
I hope the `NO_COLOR` variable is properly implemented. It doesnt seem to be _that_ standardized so I hope this is good.

johanstokking (approved):
LGTM

johanstokking:
@Jaime-Trinidad did you test this locally?

LDannijs:
(apologies for messing up the commit history. There were commits from a different branch that i had accidentally been using on here, so i tried to fix it 😓. Need to learn to double check these things)

LDannijs:
So few changes made:
- Left `macVersion` empty and removed `downlinkEncoder` and `downlinkDecoder`
- Made sure the index file in a company folder isn't overwritten if devices already exist in it.
- Added a check for the `vendorProfileID` so it increments properly.
- Also added a check if there is an available company id in the main index.yaml file.

LDannijs:
Did a few things:
- use bash instead of shell for better compatibility (linux especially)
- removed vendorProfileID check and creation (because it will be replaced with vendorIDs in #736, so wait with pushing this PR until that one is done)
- make sure device name is at least 3 characters long to comply with validation
- Add profile creation step to let user decide which ones they want to add.
- check the index.yaml file of a vendor to make sure a device with the same name isnt overwritten when adding a new one
@Jaime-Trinidad

LDannijs:
Turned the command to run the script into a make command to match the other commands.

And with #736 now being merged i can reopen this PR.

PR #702, opened 2023-11-28
new vendor and new device 

yusufscelik:
#### Summary
add new vendor "pessl_instruments" and add device from this vendor which is "iMetos LoRAIN"

#### Changes

add new vendor and the files for this vendor


yusufscelik:
Hey @Jaime-Trinidad i am sure that i run `make validate` before i made a pull request, but i am getting **"some checks were not successful"** error message. Can you maybe tell me where i made a mistake? I am also getting used to using github so because of that it can be some conflicts. Last time we added new vendor "tip" successfully but i couldnt also add this vendor.

PR #633, opened 2023-06-07
Update validation.js

Jaime-Trinidad:
#### Summary

There are more users having the `EAGAIN` error when they run `make validate`. There was a fix already made using async functions,I have been working in this possible solution, it slows the validation process.  Take a look to it

#### Changes

- Exec function usage: Switched from exec to spawn for better control.
- Asynchronous handling: Improved asynchronous operations using async and await.
- Error handling: Enhanced error management in child processes.
- Multiple child processes: Introduced delay between child process creations.
- Logging: Replaced console.debug with console.log for better logging.
- Promise handling: Improved promise management with await.
- Execution delay: Introduced an index based delay replacing setTimeout.
- Error exit: Removed process.exit(1) and incorporated error handling.

Functions modified:

- **validatePayloadCodecs**: Redesigned for better async handling, control, error management, and prevention of EAGAIN errors.
- **requireImageDecode**: Swapped exec with spawn for improved control and error handling. Errors are now propagated via promises.
- **vendors.vendors.forEach**: Refined loop with async/await for improved promise handling and graceful error management.

#### Notes for Reviewers

...

#### Release Notes

- ...

PR #726, opened 2023-12-19
Update validation.js (and comments)

LDannijs:
#### Summary

Picking up #633 to fix the comments in response to @KrishnaIyer


johanstokking:
What is the problem that we're trying to solve here?

LDannijs:
@KrishnaIyer asked me to take over that PR and to update the comments so.
Or did you mean the validate itself?

LDannijs:
Whoops that mention was an accident

johanstokking:
What is the problem this pull request is trying to solve?

LDannijs:
Running the validation very often results in an `EAGAIN` error. I do not know the details of it, only that it is a memory issue, but this fixes it by slowing the process down and other fixes (read Jaime's original PR for full details).

johanstokking:
> Running the validation very often results in an `EAGAIN` error. I do not know the details of it, only that it is a memory issue, but this fixes it by slowing the process down and other fixes (read Jaime's original PR for full details).

Do you encounter this yourself? Are there reports from users?

LDannijs:
I encounter it everytime i run it, and users also often do.

Jaime-Trinidad:
Many users seem to be encountering this issue. I believe the following code could provide a temporary solution, though it may slow down the process. It still occasionally encounters an `EAGAIN` error, and I haven't fully grasped the underlying cause yet, especially since it doesn't appear to use all the available memory on my end. Interestingly, this issue has never occurred on my VM running Ubuntu.

Additionally, it's important to note that `error` and `close` are mutually exclusive in this code. The code is designed to promise only one outcome: if an error occurs, it rejects the promise. On a successful close, the promise is resolved. However, if it closes with issues, the promise is also rejected.

johanstokking (commented):
If the issue is that too many child processes run at the same time, why not run them sequentially instead? What is the point of parallelism if they're scheduled 1 second apart?

`validatePayloadCodecs()` is awaited. So instead of pushing promises to an array and awaiting them, just await every run.

LDannijs:
So i decided to at least give it a shot and run the feedback through chatgpt and it seems it runs really well now. It's now running through it sequentially and it's not giving me any errors. Hope this is what we're looking for.

@Jaime-Trinidad please take a look as well.

bin/validate.js
```diff
@@ -113,16 +109,38 @@ function validatePayloadCodecs(vendorId, payloadEncoding) {
...
-          `bin/runscript -codec-path "${r.fileName}" -routine ${r.routine} -input '${JSON.stringify(r.input)}'`,
-          (err, stdout, stderr) => {
-            if (err) {
-              reject(err);
-            } else if (stderr) {
-              reject(stderr);
+  for (let index = 0; index < runs.length; index++) {
+    const r = runs[index];
+    await new Promise((resolve, reject) => {
+      setTimeout(async () => {
```
johanstokking, line 115: You can remove this

LDannijs: I've made a change that should hopefully be what your intention is.

johanstokking (approved):

Jaime-Trinidad (approved):

Issue #108, opened 2021-01-28
Enhance validation


johanstokking:
#### Summary

Enhance validation

#### Why do we need this?

To reduce errors and time spend on trial and error.

#### What is already there? What do you see now?

Validation works, but there are still things that can slip through.

#### What is missing? What do you want to see?

I want to make validation check more things. I also want the output to be clearer. Finally I want to be able to validate only one vendor at a time.

#### How do you propose to implement this?

1. Check casing of filenames; everything should be lowercase
2. Check if there are vendors in the `vendor` folder which are not in de `vendor/index.yaml`. This catches cases where people accidentally forgot to add the vendor entry
3. Be able to pass an argument to the validation script, to only validate a specific vendor
4. Check if the vendor has a vendor ID issued by the LoRa Alliance. If not, make sure that `vendorProfileID` in the profiles are empty (`0`), as they should be
5. LoRaWAN checks; we may want to add schema validation that the maximum EIRP does not exceed the regulatory limits. Not sure if we can make this water tight, but we can catch some common mistakes here

#### Can you do this yourself and submit a Pull Request?

Yes

neoaggelos:
Adding to the list in the description:

- Verify that the specified codec files exist (extra; and are valid Javascript)

PR #736, opened 2024-01-12
Validate features

LDannijs:
#### Summary

I stumbled upon the issues from @johanstokking  and @KrishnaIyer (#108 and #405) and decided to give it a crack. I am very curious to hear back if these implementations are correct, they would help out a lot.

#### Changes

- Johan's issue:
   - Check if there are vendors in the vendor folder which are not in `vendor/index.yaml`.
   - Ability to pass an argument to the validation script to only validate a single vendor
      - Updated README with explanation

- Krishna's issue:
   - Check if the `vendorProfileID` is unique per vendor
     (Now this checks for 2 things: is the value unique and is it missing? if it's not unique it will stop the code, and if its missing it will just log it and continue)
   - Also updated every `vendorProfileID` to match this new feature to the validation

- Own changes:
   - Updated the PR template slightly
   - Improved the `vendorProfileID` explanation a bit further to make it even clearer.

#### Notes for Reviewers

@johanstokking I had a discussion with @Jaime-Trinidad about how much we should check the `vendorProfileID` and decided to just check for uniqueness and presence. Checking it against the `vendorID` was a bit outside of the scope since not every vendor adds it.

You asked about EIRP checking as well but i heard there can be some special exceptions so i left that out for now too.


Jaime-Trinidad (approved):

LDannijs:
I've added another small feature: add a pre-commit message that runs the validation on it's own and displays the error if the validation failed:

Screenshot 2024-01-29 at 10 00 57

I sometimes see users failing to run the validation themselves so this might help to negate that instead of Jaime having to do it himself. I'm not entirely sure if this is something we'd need but I'll be happy to hear feedback.

Makefile
```diff
@@ -34,7 +36,11 @@ deps.update:
 
 .PHONY: validate
 validate:
-	$(NPM) run validate
+	@if [ -z "${vendor-id}" ]; then \
+        $(NPM) run validate; \
+    else \
+        $(NPM) run validate -- --vendor-id $(vendor-id); \
+    fi
```
johanstokking, line 43: Nit; I think you can remove the single line construct with `\` and just use multiple lines?

bin/pre-commit
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
johanstokking, line 1: I think this file can just be:

```bash
#!/bin/bash

@set -e
make validate fmt
```

package.json
```diff
@@ -6,6 +6,7 @@
     "validate": "bin/validate.js"
   },
   "scripts": {
+    "prestart": "cp bin/pre-commit .git/hooks/ && chmod +x .git/hooks/pre-commit && echo 'hook copied'",
```
johanstokking, line 9: There's no `start` nor is there a `server.js` so what's the point of `prestart`? Shouldn't this be `install-hooks` and shouldn't we mention in the README to run it?

bin/validate.js
```diff
@@ -202,6 +208,44 @@ function formatValidationErrors(errors) {
...
+// Get the list of vendor folders in the vendor directory
+const vendorFolders = fs
+  .readdirSync('./vendor', { withFileTypes: true })
+  .filter((dirent) => dirent.isDirectory())
+  .map((dirent) => dirent.name);
+
+// Check for vendors in the vendor folder that are not in the vendor/index.yaml
+const vendorsNotInIndex = vendorFolders.filter((vendorId) => {
+  // Check if any vendor ID in the index is present in the folder name, or vice versa
+  const normalizedVendorId = vendorId.toLowerCase(); // Normalize for case-insensitive comparison
```
johanstokking, line 232: Remove; this is case sensitive

bin/validate.js
```diff
@@ -202,6 +208,44 @@ function formatValidationErrors(errors) {
...
+  return (
+    !vendorIdsInIndex.some((indexId) => normalizedVendorId.includes(indexId.toLowerCase())) &&
+    !vendorIdsInIndex.some((indexId) => indexId.toLowerCase().includes(normalizedVendorId))
```
johanstokking, line 235: Keep everything case sensitive

bin/validate.js
```diff
@@ -273,41 +328,71 @@ vendors.vendors.forEach((v) => {
...
-          if (regionProfile.codec && !codecs[regionProfile.codec]) {
-            const codec = yaml.load(fs.readFileSync(`${folder}/${regionProfile.codec}.yaml`));
-            if (!validateEndDevicePayloadCodec(codec)) {
-              console.error(
-                `${key}: codec ${regionProfile.codec} invalid: ${formatValidationErrors(
-                  validateEndDevicePayloadCodec.errors
-                )}`
+
+            // Check if vendorProfileID is unique within the same company folder
+            if (profile.vendorProfileID !== undefined) {
```
johanstokking, line 340: `0` should act the same as `undefined`

bin/validate.js
```diff
@@ -273,41 +328,71 @@ vendors.vendors.forEach((v) => {
...
+                  `${key}: profile ${vendorID}/${regionProfile.id} invalid: ${formatValidationErrors(
+                    validateEndDeviceProfile.errors
+                  )}`
+                );
+                process.exit(1);
+              }
+
+              console.log(`${key}: profile ${vendorID}/${regionProfile.id} valid`);
+            } else {
+              console.warn(`${key}: vendorProfileID is missing in ${vendorID}/${regionProfile.id}.yaml`);
```
johanstokking, line 374: This is not a warning

vendor/abeeway/abeeway-compact-tracker-pro-eu868.yaml
```diff
@@ -1,7 +1,7 @@
 # Vendor profile ID, can be freely issued by the vendor
 # This vendor profile ID is also used on the QR code for LoRaWAN devices, see
 # https://lora-alliance.org/sites/default/files/2020-10/LoRa_Alliance_Vendor_ID_for_QR_Code.pdf
-vendorProfileID: 0
+vendorProfileID: 1
```
johanstokking, line 4: Revert all changes when it was `0`

Makefile
LDannijs: i can't seem to figure out how to make it multiple lines without getting errors, so ill make it a single line

package.json
LDannijs: I will instead add install-hooks to `make deps`, makes it easier for users.

bin/validate.js
LDannijs: But isnt 0 a valid value?

Makefile
```diff
@@ -19,6 +19,8 @@ GO = go
 GOBIN = $(PWD)/bin
 export GOBIN
 
+vendor-id=
```
johanstokking, line 22: Since we are also using other variables in this file (see above this line), I think it's better to do something like this:

```
VENDOR_ID=
```

and then document as:

```bash
VENDOR_ID=myvendor make validate
```

Makefile
```diff
@@ -37,7 +37,7 @@ deps.update:
 
 .PHONY: validate
 validate:
-	@if [ -z "${vendor-id}" ]; then $(NPM) run validate; else $(NPM) run validate -- --vendor-id $(vendor-id); fi
+	@if [ -z "${VENDOR_ID}" ]; then $(NPM) run validate; else $(NPM) run validate -- --VENDOR_ID $(VENDOR_ID); fi
```
johanstokking, line 40: The script argument is still `vendor-id`

LDannijs: Huh no i just changed it to `VENDOR_ID`?

johanstokking: Let me rephrase: the script argument should still be `vendor-id`, that is idiomic for flags.

LDannijs: Ok i *think* i understand it now. Pushed another fix.

johanstokking (commented):
This now implements one of the options mentioned in https://github.com/TheThingsNetwork/lorawan-devices/issues/405#issuecomment-1278960620.

The other option is that a `vendorProfileID` refers to a unique combination of device ID + firmware version + hardware version + region. When we have that, we have the profile and the codec, but we also know exactly which device it is.

So I'm suggesting the following change:

- Remove `vendorProfileID` from the profile definition
- Add a new section in the vendor's index file (not the vendor index but the index of the vendor), top-level (same as `endDevices`), called `profileIDs`
- This new section is a key/value mapping of vendor profile ID (numeric) to either an object with device ID + hardware version + firmware version + region, or a profile ID + codec ID

Example diff:

```diff
diff --git a/vendor/the-things-industries/index.yaml b/vendor/the-things-industries/index.yaml
index e1a0d5769..7f0366100 100644
--- a/vendor/the-things-industries/index.yaml
+++ b/vendor/the-things-industries/index.yaml
@@ -1,2 +1,8 @@
 endDevices:
   - generic-node-sensor-edition
+profileIDs:
+  '42':
+    endDeviceID: 'generic-node-sensor-edition'
+    firmwareVersion: '1.0'
+    hardwareVersion: '1.1'
+    region: 'EU863-870'
```

This requires some JSON Schema adaptations. Something like:

```diff
diff --git a/schema.json b/schema.json
index c5d3afcba..9e1f4f5a3 100644
--- a/schema.json
+++ b/schema.json
@@ -79,6 +79,48 @@
           "items": {
             "$ref": "#/definitions/identifier"
           }
+        },
+        "profileIDs": {
+          "type": "object",
+          "patternProperties": {
+            "^[0-9]+$": {
+              "oneOf": [
+                {
+                  "type": "object",
+                  "properties": {
+                    "endDeviceID": {
+                      "$ref": "#/definitions/identifier"
+                    },
+                    "hardwareVersion": {
+                      "type": "string",
+                      "minLength": 1
+                    },
+                    "firmwareVersion": {
+                      "type": "string",
+                      "minLength": 1
+                    },
+                    "region": {
+                      "$ref": "#/definitions/region"
+                    }
+                  },
+                  "required": ["endDeviceID", "hardwareVersion", "firmwareVersion", "region"]
+                },
+                {
+                  "type": "object",
+                  "properties": {
+                    "id": {
+                      "$ref": "#/definitions/identifier"
+                    },
+                    "codec": {
+                      "$ref": "#/definitions/identifier"
+                    }
+                  },
+                  "required": ["id", "codec"]
+                }
+              ]
+            },
+            "additionalProperties": false
+          }
         }
       },
       "additionalProperties": false
```

LDannijs:
I'm aware not everything is properly implemented, had to fight my script a bit (which will be removed), the schema is missing hardwareVersion, some files are missing proper formatting, yadayada. Will continue this later.

LDannijs:
@johanstokking i am pretty sure the implementation will have some problems but just as check if this is the correct direction.

README.md
```diff
@@ -161,6 +161,16 @@ All vendor data is referenced from the **Vendor device index** file: `vendor/<ve
 endDevices:
   - device-a
   - device-b
+# The profileIDs is a distinct (preferably incremental) value for every unique profile listed in the vendor's folder.
```
johanstokking, line 164: Progressive insight: there's actually no reason for it to be incremental. If vendors want to use ranges for regions because they think that's handy, they should. I know you copied this but please update.

README.md
```diff
@@ -161,6 +161,16 @@ All vendor data is referenced from the **Vendor device index** file: `vendor/<ve
...
+# This value can be freely issued by the vendor and is also used on the QR code for LoRaWAN devices, see
+# https://lora-alliance.org/wp-content/uploads/2020/11/TR005_LoRaWAN_Device_Identification_QR_Codes.pdf#page=8
+# NOTE: The profileIDs is different from the vendorID.
+profileIDs:
+  '1':
+    endDeviceID: 'device-a'
+    firmwareVersion: '1.0'
+    hardwareVersion: '1.0'
+    region: 'EU863-870'
```
johanstokking, line 173: This only explains one option: the fully qualified end device's region. The other option (only profile ID and codec ID) should also be documented here.

bin/script.js
```diff
@@ -0,0 +1,104 @@
+#!/usr/bin/env node
```
johanstokking: What is `script.js` supposed to do?

bin/validate.js
```diff
@@ -208,17 +208,10 @@ function formatValidationErrors(errors) {
...
-if (!validateVendorsIndex(vendorsIndex)) {
-  console.error(`${options.vendor} index is invalid: ${formatValidationErrors(validateVendorsIndex.errors)}`);
-  process.exit(1);
-}
-console.log(`vendor index: valid`);
+const vendors = yaml.load(fs.readFileSync(options.vendor));
 
 // Get the list of vendor IDs from the vendor/index.yaml
-const vendorIdsInIndex = vendorsIndex.vendors.map((vendor) => vendor.id);
+const vendorid = vendors.vendors.map((vendor) => vendor.id);
```
johanstokking, line 214: nit;

```suggestion
const vendorId = vendors.vendors.map((vendor) => vendor.id);
```

schema.json
```diff
@@ -545,40 +586,7 @@
       "title": "End Device Profile",
       "description": "End device capabilities and boot parameters",
       "type": "object",
-      "examples": [
```
johanstokking, line 548: Why remove?

bin/script.js
```diff
@@ -0,0 +1,104 @@
...
+
+      // Assuming firmwareVersion and hardwareVersion are correctly extracted
+      // Iterate through firmware versions to access profiles
+      endDevice.firmwareVersions.forEach((version) => {
+        Object.keys(version.profiles).forEach(async (region) => {
+          const profilePath = `./vendor/${v.id}/${version.profiles[region].id}.yaml`;
+          const profile = yaml.load(fs.readFileSync(profilePath));
+
+          // Use '0' as a fallback but decide not to proceed if it is '0'
+          let profileID = profile.vendorProfileID || '0';
```
johanstokking: You should convert this to a number (and not treat it as a string). This is because `'0002'` doesn't equal `'2'` while 2 equals 2.

bin/script.js
LDannijs: It was just used to put the new `profileIDs` field in the indexes with all the information. I removed it since it was only for single use.

schema.json
LDannijs: Thought it wasn't needed since vendorprofileid is gone but im stupid and didnt realize the other properties still apply 😇

bin/validate.js
LDannijs: I can't cause of the code below:
```
const vendorsNotInIndex = vendorFolders.filter((vendorId) => {
  // Check if any vendor ID in the index is present in the folder name, or vice versa
  return (
    !vendorid.some((indexId) => vendorId.includes(indexId)) && !vendorid.some((indexId) => indexId.includes(vendorId))
  );
});
```
But i will change it to `const vendorIds`

vendor/adeunis/index.yaml
```diff
@@ -1,22 +1,54 @@
...
+  '5':
+    endDeviceID: temp
+    firmwareVersion: '1.0'
+    hardwareVersion: '1.00'
+    region: EU863-870
+  '6':
+    endDeviceID: pulse
+    firmwareVersion: '1.0'
+    hardwareVersion: '1.0'
+    region: US902-928
```
johanstokking, line 51: So just for my understanding; you issued these numbers, right? They were not in the DR before?

LDannijs: For Adeunis specifically, yes. Reason being if you look at the files i changes all the profiles had `16` as the `vendorProfileID`, but we cannot use the same number for multiple profiles with the new `profileIDs`

johanstokking: I see, but we should not make profile IDs up for them. Vendors own these things. If they suddenly start from 1, it conflicts with what we already have.

LDannijs: I understand, how do you suggest we fix that?

johanstokking: Yes

LDannijs: I get that, but what about adeunis specifically for example? They have multiple profiles with the same `vendorProfileID`.

LDannijs:
Alright @johanstokking think I finally understood it now (apologies)
I now removed the `profileIDs` from vendors where `vendorProfileID` was a repeating id, a vendorid, a made up number, etc. 

I think I messed up cause, yes, I initially created a check for if the `vendorProfileID` was unique, but I rolled back that change.

vendor/decentlab/profile-as923.yaml
```diff
@@ -1,4 +1,3 @@
-vendorProfileID: 3
```
johanstokking, line 1: This profile is used by many Decentlab devices right?

vendor/decentlab/index.yaml
```diff
@@ -44,3 +44,24 @@ endDevices:
...
+  '2':
+    endDeviceID: dl-zn2
+    firmwareVersion: '1.5.71'
+    hardwareVersion: '1'
+    region: US902-928
+  '3':
+    endDeviceID: dl-zn2
+    firmwareVersion: '1.5.71'
+    hardwareVersion: '1'
+    region: AS923
```
johanstokking, line 62: But this references exactly one device

In this case, there should simply be a profile and codec ID be defined. If there are multiple different codecs used, the number should not be defined in this file.

LDannijs: Yeah it comes up often that its a single profile used by multiple devices but also multiple codecs. So how do i approach the numbering? you say `should not be defined in this file` but can you give me an example?

bin/validate.js
```diff
@@ -204,6 +210,34 @@ function formatValidationErrors(errors) {
...
+const vendorFolders = fs
+  .readdirSync('./vendor', { withFileTypes: true })
+  .filter((dirent) => dirent.isDirectory())
+  .map((dirent) => dirent.name);
+
+// Check for vendors in the vendor folder that are not in the vendor/index.yaml
+const vendorsNotInIndex = vendorFolders.filter((vendorId) => {
+  // Check if any vendor ID in the index is present in the folder name, or vice versa
+  return (
+    !vendorIds.some((indexId) => vendorId.includes(indexId)) && !vendorIds.some((indexId) => indexId.includes(vendorId))
```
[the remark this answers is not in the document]
LDannijs: Think I fixed it. Wasn't properly validating before, whoops.

Jaime-Trinidad (approved):

Issue #745, opened 2024-02-06
Generate devices csv


benolayinka:
#### Summary

Can we generate a flat list of devices as a csv? We can then sync this to airtable

#### Why do we need this?

Sync data to airtable

#### What is already there? What do you see now?

Folder hierarchy

#### What is missing? What do you want to see?

name, vendor, description, image link, sensors, radios, height, width, length, weight, battery?, ip rating, urls, mac version?, rp version?, supports class b / class c?,

johanstokking:
Most of this is quite straight forward and is readily available per end device.

The data model has a hierarchy:

- Vendor
  - Profile
  - End device
    - Hardware version
    - Firmware version (links to one or more supported hardware versions and a profile for each region)

The MAC version, RP version and LoRaWAN classes are in the profile. There may be differences between regions, and there can be devices with multiple firmware versions so we need to add some heuristics here.

For example, we could take the last firmware version and the first profile to determine these fields.

LDannijs:
Iirc i wasn't able to change the labels before but i will remember it for next time.

LDannijs:
PR #749 was merged, closing this.

PR #749, opened 2024-02-14
Generate CSV for Partner Portal

LDannijs:
#### Summary

Generate a csv file with information regarding all the devices in the Device Repository for use in the Partner Portal. Issue #745 @benolayinka 

image

#### Changes

- add a javascript script for generating a csv
- Remove example from Atim (Seems to be the same as [Windsensor example](https://github.com/TheThingsNetwork/lorawan-devices/blob/master/vendor/example/windsensor.yaml), so i say remove it)

#### Checklist for Reviewers

- [ ] `vendorProfileID` should not be `vendorID`.

#### Notes for Reviewers

- @johanstokking The script grabs the highest macversion number of all the profiles the device associates with. This might need to be changed but with a device sometimes having tens of profiles i found this to be the best way to consolidate it.

#### Release Notes

- ...


csv.js
```diff
@@ -0,0 +1,86 @@
+const fs = require('fs');
```
johanstokking, line 1: This file should go in `bin`

csv.js
```diff
@@ -0,0 +1,86 @@
...
+// Function to extract data from device.yaml and create a CSV line
+const extractData = (filePath, vendor) => {
+    try {
+        const fileContents = fs.readFileSync(filePath, 'utf8');
+        const data = yaml.load(fileContents);
+        if (data && data.name && data.description) { //only read files with these in its content
+            const name = data.name;
+            const description = data.description.replace(/"/g, "'");
+            const sensors = Array.isArray(data.sensors) ? `"${data.sensors.join(', ')}"` : '';
+            const imageUrl = data.photos?.main ? `"${baseUrl}/${vendor}/${data.photos.main}"` : '';
```
johanstokking, line 16: At some point we have to sync this to a cloud bucket (like S3) with content delivery network (like CloudFront) and resize images in certain fixed dimensions. Some of the photos here are very big, as they should be because that allows us to do high res stuff with them.

csv.js
```diff
@@ -0,0 +1,86 @@
...
+            data.firmwareVersions.forEach(firmwareVersion => {
+                Object.values(firmwareVersion.profiles).forEach(profile => {
+                    const relatedFilePath = path.join(path.dirname(filePath), `${profile.id}.yaml`);
+                    try {
+                        const profileContents = fs.readFileSync(relatedFilePath, 'utf8');
+                        const profileData = yaml.load(profileContents);
+                        const macVersion = profileData.macVersion || '';
+            
+                        // Compare and update the highest macVersion found
+                        if (macVersion.localeCompare(highestMacVersion, undefined, { numeric: true, sensitivity: 'base' }) > 0) {
```
johanstokking, line 41: I think this is pretty acceptable yes

csv.js
LDannijs: So.. is there anything you want me to do with it? Sorry if i'm not getting the request clear.

johanstokking: That depends on what will happen with these photos. We shouldn't render them as-is: they're too big (and they should be that big) and GitHub raw's URL is too unreliable to have nothing between the user's browser and this URL.

I don't have a request for change for this PR, but it may need to be implemented later and then this script will need the update.

LDannijs: I am also not sure myself how these will be rendered, but i found the raw github url to be the best for now.
Best to hear from @benolayinka if it's the right method.

benolayinka: for now it's fine as you have it, at least for testing. in the future we will need to resize images and link to the web-friendly size.

benolayinka:
As far as resizing images, I think @mjamescompton is using Hugo in this repo to build the website and might already be doing some image resizing

benolayinka:
@LDannijs it looks like there are duplicate devices in the csv, is this maybe because it's generating one per profile or something? we need at least one unique identifier per device

@adriansmares I also broke the table at that API endpoint testing w dummy data, so I'll have to update the secret before we merge this

Screenshot 2024-02-22 at 11 26 08

LDannijs:
Just took a look and it seems some devices are named differently in filename but have the same name/description. Minor changes usually, for example the AM319 has an AM319L model without lcd, but is named AM319 in device name. I'll see if i can update the titles of these devices.

johanstokking:
@LDannijs probably good to add a validation for the uniqueness of the name then.

The namespace should however not be global; there can be multiple "LoRaWAN street light module" from different vendors.

What do we do with resizing images? Can we let the partner portal frontend do it? I'd prefer that anyway than doing it here.

LDannijs:
@johanstokking good idea. I will add that to #736

LDannijs:
Alright this should fix it @benolayinka, can you check if the validation looks good @johanstokking?

bin/validate.js
```diff
@@ -51,6 +51,25 @@ function requireFile(path) {
...
+        if (deviceNames.has(deviceData.name)) {
+          throw new Error(`Duplicate device name found: ${deviceData.name} in folder ${folder}`);
+        }
+        deviceNames.add(deviceData.name);
+      } else {
+        console.error(`Error: Unable to load device data from file ${file}, or 'name' property is missing.`);
+      }
+    }
+  }
+}
```
johanstokking, line 71: Isn't this better to just make part of the loop below?

johanstokking (approved):

bin/validate.js
```diff
@@ -215,14 +216,42 @@ const vendorProfiles = {};
 vendors.vendors.forEach((v) => {
   const key = v.id;
   const folder = `./vendor/${v.id}`;
+  const vendorIndexPath = `${folder}/index.yaml`;
+
+  // Check for the existence of the vendor's index.yaml file
+  if (!fs.existsSync(vendorIndexPath)) {
+    return; // Skip this vendor if the index file doesn't exist
+  }
```
johanstokking, line 224: This is the `ENOENT` check below (L257). It's probably cleaner to do a `statSync()` here with the `ENOENT` check like below. It can be removed from below, which removes an indentation level too.

bin/validate.js
```diff
@@ -215,14 +216,42 @@ const vendorProfiles = {};
...
+
+  const vendorIndex = yaml.load(fs.readFileSync(vendorIndexPath, 'utf8'));
```
johanstokking, line 226: Duplicate of L264

bin/validate.js
```diff
@@ -215,14 +216,42 @@ const vendorProfiles = {};
...
+  // Initialize a new Set for each vendor to track device names, ensuring isolation
+  const deviceNames = new Map();
```
johanstokking, line 228: Why `Map` if all you need is `{}` like `codecs`

bin/validate.js
```diff
@@ -215,14 +216,42 @@ const vendorProfiles = {};
...
+
+  vendorIndex.endDevices?.forEach((deviceFileName) => {
```
johanstokking, line 230: There is already a loop below for each end device, why have another loop here?

bin/validate.js
```diff
@@ -215,14 +216,42 @@ const vendorProfiles = {};
...
+    const deviceFilePath = `${folder}/${deviceFileName}.yaml`;
+
+    if (!fs.existsSync(deviceFilePath)) {
+      return;
+    }
+
+    const deviceData = yaml.load(fs.readFileSync(deviceFilePath, 'utf8'));
+    const existingFilePath = deviceNames.get(deviceData.name);
```
johanstokking, line 238: This also happens on L247, why duplicate?

benolayinka:
@LDannijs moving csv.js to /bin breaks the makefile, and running node bin/csv.js throws an error because there is a relative ../vendor path somewhere. Can you please update csv.js so that it can be run from the makefile as node bin/csv.js, since that is how it will be run in production?

benolayinka:
@LDannijs I tested the latest push and there is still one duplicate, "Flood Sensor"

Screenshot 2024-02-27 at 12 04 41

vendor/ambitlocker/ambitlocker.yaml
```diff
@@ -1,4 +1,4 @@
-name: Ambitlocker lock
+name: The lock
```
johanstokking, line 1:
```suggestion
name: Lock
```

vendor/aquascope/wwd.yaml
```diff
@@ -1,4 +1,4 @@
-name: Flood Sensor
+name: FLOLWE01 - Flood Sensor
```
johanstokking, line 1: Why?

vendor/acrios/acr-slc-100l-zhaga-dali.yaml
```diff
@@ -1,4 +1,4 @@
-name: LoRaWAN street light module
+name: LoRaWAN street light module (ZHAGA DALI)
```
johanstokking, line 1: Why?

vendor/browan/tbsl100.yaml
```diff
@@ -1,4 +1,4 @@
-name: Sound Level Sensor
+name: TBSL100 - Sound Level Sensor
```
johanstokking, line 1: Why?

vendor/aquascope/wwd.yaml
LDannijs: Cause theres multiple devices named flood sensor, and ben is getting a duplicate error: https://github.com/TheThingsNetwork/lorawan-devices/pull/749#issuecomment-1966305882

johanstokking: The uniqueness should be on vendor name + device name, not just device name

LDannijs: Ok i understand, but then we have a problem. If you look at that comment from Ben you can see that airtable can not link the name and vendor together, so it will count devices with the same name as a duplicate. I dont know if you can fix this in airtable somehow @benolayinka but we need to figure out how we're going to do that.

benolayinka:
Yeah, I need a single unique identifying field. You could generate an ID as a hash of the name and vendor or something.

vendor/acrios/acr-slc-100l-zhaga-dali.yaml
LDannijs: This one we do need cause the device `acr-slc.yaml` (same vendor) is already called `LoRaWAN street light module`, so since this device uses the zhaga dali "protocol" it needs to be different.

LDannijs:
@benolayinka added an id field thats just device name + vendor, should do the trick. Rolled back some of those name changes too.

vendor/aquascope/wwd.yaml
johanstokking: How about adding simply another column with the full name, just for enforcing uniqueness?

LDannijs: Made a column with device name + vendor, that way we can be fully sure its unique.

vendor/hbi/hbi-mla20-3l0606c.yaml
```diff
@@ -1,4 +1,4 @@
-name: HBI emergency lighting series 300 eagle
```
johanstokking, line 1: Please make sure the name starts with a capital.

Probably easiest is to find in files with the right filter (`vendor/**/*.yaml`), case sensitive and regex enabled for `name: [a-z]`

LDannijs: And what about this for example?
https://github.com/TheThingsNetwork/lorawan-devices/blob/8cd4bbf6dc318e23f1c6ec1cfa89e8ba42704537/vendor/tektelic/t0006923-edoctor.yaml#L1

The device is officially called `eDoctor` with a small e and capital D. Do we just ignore their naming convention or?

johanstokking: Yes those are fine

LDannijs:
So i noticed Dragino has some devices missing in their index.yaml file, which doesn't make them show up on the site/stack, but i tried to fix them yet they have codec issues that i unfortunately cannot fix myself.

johanstokking:
Thanks, this now looks good to me.

KrishnaIyer:
@LDannijs: for future PRs use the format `Closes #<issue number>` in your description. This will automatically close the issue when the PR is merged.

johanstokking:
@Jaime-Trinidad can you review so that we can merge this?

Jaime-Trinidad (approved):

Jaime-Trinidad:
> @Jaime-Trinidad can you review so that we can merge this?

I reviewed and tested it, devices.csv looks fine.

PR #761, opened 2024-03-14
Added TEKTELIC Stork and Breeze Devices

ssharma-tektelic:
#### Summary

Added TEKTELIC Stork and Breeze devices.
#### Changes

Modified existing TEKTELIC files to include addition of Stork and Breeze devices, and created new codecs and profiles for these new devices.

#### Checklist for Reviewers

- [ ] `vendorProfileID` should not be `vendorID`.

#### Notes for Reviewers

...

#### Release Notes

- ...


bin/validate.js
```diff
@@ -59,7 +59,9 @@ async function requireDimensions(path) {
         reject(new Error(`load image ${path}: ${err}`));
```
johanstokking, line 59: Please revert changes in this file

johanstokking (commented):
Thanks, one comment

bin/validate.js
ssharma-tektelic: I've reverted the changes in the file but it still shows the last line as modified, are you able to change it?

vendor/tektelic/t0007806-864-profile.yaml
```diff
@@ -0,0 +1,8 @@
+vendorProfileID: 415
```
johanstokking, line 1: @ssharma-tektelic We have some work in progress (https://github.com/TheThingsNetwork/lorawan-devices/pull/736) that's gonna change how the profile ID works.

I see that you're reusing the same profile ID for multiple different profiles. This will no longer be possible.

Is `415` the actual number you use, or does it happen to be your vendor ID?

Please either use unique numbers for each profile or unset this field.

johanstokking (commented):
You should be able to resolve the diff in `bin/validate.js` by adding an empty line in the bottom.

vendor/tektelic/t0007806-864-profile.yaml
ssharma-tektelic: 415 is the vendor ID I found in vendor/index.yaml. If I unset it, do I remove the line completely or leave it as "vendorProfileID:"?

johanstokking: @ssharma-tektelic remove the line please

johanstokking (approved):

johanstokking (approved):

PR #772, opened 2024-04-08
Fix missing sensors and add device types

### LINKED ISSUE #769: Missing Sensors on Devices
#### Summary

There are 252 devices without sensors, it is needed to review if there is the need to add sensors that are not in `.yaml` file or contact device makers so they can add them. Also this will work for the dropdown menu

#### What do you see now?

252 devices without sensors in `.yaml` file

i.e. Elvaco `cmi4130.yaml` they deleted `sensors` 

#### What do you want to see instead?

Reduce the number of devices without this information

...

@LDannijs let's work on this.

LDannijs:
#### Summary

Looked through the devices csv to note which devices are missing sensors. Was able to add it for a lot but Netvox, Radio Bridge and Tektelic have too many to manually change. Also added `devicetype` to compensate for devices with no sensors. Closes #769 

#### Changes

- Add missing sensors for a lot of devices missing them
  - EXCEPT Netvox, Radio Bridge and Tektelic (too many devices) and devices with no sensors (devicetype)
- Add `devicetype` to the schema to identify what type of device it is 
  - Some devices that don't have sensors now have `devicetype` to be able to filter for them.
- Add `devicetype` to the csv script
- Make images mandatory to add to device file
- Remove `The-box development` and `Grinn` 
  - No images for their devices, barren and old info, no information to be found about them online
- fix Elvaco images

#### Notes for Reviewers

@Jaime-Trinidad lets discuss if this is correct and if there are things we're missing.


LDannijs:
Decided to update Netvox, Radio bridge and Tektelic anyway :)

schema.json
```diff
@@ -918,6 +918,12 @@
           "description": "Instructions for finding firmware version",
           "examples": ["In the WebFig configuration panel, the firmware version is stated in the top left (e.g RouterOS v6.47.9)"]
         },
+        "devicetype": {
+          "type": "array",
+          "items": {
+            "enum": ["sensor", "devboard", "actuator", "module", "converter", "coveragetest", "display"]
```
johanstokking, line 924: The properties are not really mutually exclusive and collectively exhaustive anymore this way.

For example:
- If the device has any sensor in `sensors`, it's a sensor
- Can it be a dev board and a module or are they mutually exclusive? Isn't a device typically either a commercial of the shelf device, a dev board or a module?
- What does converter mean? Isn't this what `bridgeInterfaces` is for?

schema.json
```diff
@@ -936,15 +942,18 @@
               "conductivity",
               "current",
               "digital input",
+              "digital output",
```
johanstokking, line 945: A digital output is not a sensor

If needed, we can add a property `actuators`

schema.json
johanstokking: I understand this is useful for the CSV export, but I think we should deduce this from the other properties instead of declaring a new one here.

LDannijs: I had this discussion with @Jaime-Trinidad, and he suggested something to categorize the types of devices available. Not every device has sensors (like actuators) so i felt this would help the filtering (when that is eventually implemented).

In my opinion, the more we can differentiate the better. Jaime told me users search a lot for devboards, so categorizing them separately would help i imagine.

And yes `bridgeInterfaces` does that task but its not a category, users wont be able to filter by those.

johanstokking: What I mean to say is that there is a difference between what is defined in the model and what is presented to users.

Now, presentation layer requirements are applied to the data layer, requiring additional fields to be set, which may feel like redundant and duplicate and therefore easy to miss.

What we should do instead, is wherever we need it – that is, in the representation (CSV export) or presentation layer (that uses the CSV) – deduce fields from whatever is in the data repository here.

For example, if there are any `bridgeInterfaces`, it's an `actuator`.

I'm supporting adding stuff to the schema that we cannot declare otherwise (like whether it's a module, a dev board or a commercial off the shelf device, which is mutually exclusive). But other stuff, like whether it's a sensor or an actuator or a converter should be deduced. And probably, a `display` is just a capability or even an actuator, not a device type.

LDannijs: Alright i see. I just imagine it would be more work for the eventual website filtering to, for example, search for "if there are any `bridgeInterfaces`, it's an `actuator`."

So what you're suggesting is leaving module and devboard? And 

> or a commercial off the shelf device

To what products would that apply? isn't that just everything else?

johanstokking: > Alright i see. I just imagine it would be more work for the eventual website filtering to, for example, search for "if there are any `bridgeInterfaces`, it's an `actuator`."

It is much more work for device makers to fill out redundant fields than it is for some basic filtering. If you really want to do this in the CSV, why wouldn't you check `Array.isArray(d.bridgeInterfaces) && d.bridgeInterfaces.length > 0` which returns a boolean that you can export as `converter`.

> So what you're suggesting is leaving module and devboard? And
> 
> > or a commercial off the shelf device
> 
> To what products would that apply? isn't that just everything else?

Yeah I think there's `developer kit`, `module` and `cots`. That's not an array of enum but just an enum.

> EDIT: also, display is a device type, see: https://tektelic.com/products/sensors/memo-multi-purpose-display-tablet/ this one is in the repo

I don't understand what the difference is between a display vs a light vs a digital output. It all falls in the category actuators.

bin/csv.js
```diff
@@ -37,6 +37,7 @@ const extractData = (filePath, vendor) => {
       const name = data.name;
       const vendorname = vendorNamesMap.get(vendor) || vendor; // Fallback to the vendor ID if no name is found
       const description = data.description.replace(/"/g, "'");
+      const devicetype = data.devicetype || '';
```
johanstokking, line 40: Please rename to `deviceType`, lowerCamelCase

johanstokking: This applies to what is in the device repository and schema as well

LDannijs: Whoops apologies 😁

johanstokking (approved):

Jaime-Trinidad (approved):

Jaime-Trinidad (approved):

Jaime-Trinidad (approved):

PR #782, opened 2024-05-03
tlm-lw new codec adde and lkm-lw new product added

ali-ugur:
tlm-lw new codec adde and lkm-lw new product added

...

#### Changes

- ...
- ...

#### Checklist for Reviewers

- [ ] Title and description should be descriptive (Not just a serial number for example).
- [ ] `profileIDs` should not be `vendorID` and should be a unique value for every profile.
- [ ] All devices should be listed in the vendor's `index.yaml` file.
- [ ] At least 1 image per device and should be transparent.

#### Notes for Reviewers

...

#### Release Notes

- ...

PR #783, opened 2024-05-03
Fix vendorProfileID validation

LDannijs:
#### Summary

I noticed with #782 that the validation wasn't catching if `vendorProfileID` existed. I am not sure why, since it can clearly read the value, but replacing it with catching the string does the trick (and I tested it, it works)

Also they tried to change the firmware versions, which gave us a lot of problems not long ago with Dragino, so updating the pull request template too.

#### Changes

- change `if (profile.vendorProfileID)` to `if ('vendorProfileID' in profile)`
- Update pull request template with firmware versions

### DIFF
--- a/bin/validate.js
+++ b/bin/validate.js
@@ -342,7 +342,7 @@ vendors.vendors.forEach((v) => {
 
           if (!vendorProfiles[vendorID][regionProfile.id]) {
             const profile = yaml.load(fs.readFileSync(`./vendor/${vendorID}/${regionProfile.id}.yaml`));
-            if (profile.vendorProfileID) {
+            if ('vendorProfileID' in profile) {
               console.log(`\n${key}: vendorProfileID exists. This method has been replaced with VendorIDs, see:`);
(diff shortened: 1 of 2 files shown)


bin/validate.js
```diff
@@ -342,7 +342,7 @@ vendors.vendors.forEach((v) => {
 
           if (!vendorProfiles[vendorID][regionProfile.id]) {
             const profile = yaml.load(fs.readFileSync(`./vendor/${vendorID}/${regionProfile.id}.yaml`));
-            if (profile.vendorProfileID) {
+            if ('vendorProfileID' in profile) {
```
johanstokking, line 345: This is because the value `0` is falsy in Javascript. Your fix checks explicitly if the field is in the object which is better. 

The more idiomatic way if doing this is `if (typeof profile.vendorProfileID !== 'undefined')`. Technically, it is also okay if the value is undefined. However, that will never happen when reading from YAML, so it's only more idiomatic.

johanstokking (approved):

bin/validate.js
LDannijs: Ahh i see, thanks! Will keep it in mind.

PR #800, opened 2024-07-08
Extend normalized payload

LDannijs:
#### Summary

Extend the normalized payload function to 20 devices.

#### Changes

- add a normalized payload and output to 20 of the most popular devices
- Add Dragino LWL03A

#### Checklist for Reviewers

- [x] Title and description should be descriptive (Not just a serial number for example).
- [x] `profileIDs` should not be `vendorID` and should be a unique value for every profile.
- [x] All devices should be listed in the vendor's `index.yaml` file.
- [x] Firmware versions can not be changed.
- [x] At least 1 image per device and should be transparent.


lib/payload.json
```diff
@@ -137,6 +166,44 @@
...
+          "type": "object",
+          "properties": {
+            "latitude": {
+              "description": "Horizontal distance from equator",
+              "$ref": "#/definitions/latitude"
+            },
+            "longitude": {
+              "description": "Vertical distance from prime meridian",
+              "$ref": "#/definitions/longitude"
+            }
```
johanstokking, line 204: I don't think they'll be reused so you can just declare them here.

The reason why some other quantities (like percentage) are defined top-level is because they're easily reused in other places.

lib/payload.json
```diff
@@ -103,6 +128,10 @@
               "description": "Air temperature (°C)",
               "$ref": "#/definitions/temperature"
             },
+            "temperatureEXT": {
```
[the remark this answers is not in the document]
LDannijs: External sensor. Some devices have the ability to attach an external probe to measure temperature.

lib/payload.json
```diff
@@ -137,6 +166,44 @@
             }
           },
           "additionalProperties": false
+        },
+        "water": {
+          "type": "object",
+          "properties": {
+            "leak": {
+              "description": "Water leaking (string)",
```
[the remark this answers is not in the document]
LDannijs: Since the device that uses it outputs it as a string not a boolean:

https://github.com/TheThingsNetwork/lorawan-devices/blob/a15feaa9341c216d7862814415eda6a8a34bacd8/vendor/dragino/lwl03a-codec.yaml#L9-L13

lib/payload.json
LDannijs: Alright i guess i got confused there. I followed the schema as it was so i assumed every object needed to be defined top-level. Will change it 👍

lib/payload.json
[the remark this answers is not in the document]
johanstokking: It's the job of the normalizer to normalize that.

lib/payload.json
```diff
@@ -137,6 +166,44 @@
...
+              "$ref": "#/definitions/leak"
+            }
+          },
+          "additionalProperties": false
+        },
+        "action": {
+          "type": "object",
+          "properties": {
+            "motionState": {
+              "description": "State of Motion (number)",
```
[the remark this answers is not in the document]
johanstokking: See above, that is an enum state.

lib/payload.json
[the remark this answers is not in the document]
LDannijs: Oh right 😅

lib/payload.json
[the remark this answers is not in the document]
johanstokking: That will be a separate value then. Probably best to have an object `motion` with the (current) state and count.

lib/payload.json
```diff
@@ -212,9 +212,19 @@
...
           "properties": {
-            "motionState": {
-              "type": "number",
-              "description": "State of Motion (number)"
+            "motion": {
+              "type": "object",
+              "properties": {
+                "motionState": {
+                  "type": "boolean",
+                  "description": "State of Motion"
```
johanstokking, line 220:
```suggestion
                  "description": "Motion detected"
```

lib/payload.json
```diff
@@ -212,9 +212,19 @@
         "action": {
           "type": "object",
           "properties": {
-            "motionState": {
-              "type": "number",
-              "description": "State of Motion (number)"
+            "motion": {
+              "type": "object",
+              "properties": {
+                "motionState": {
```
johanstokking, line 218:
```suggestion
                "detected": {
```

We're already in `motion` so no need to have `motion.motionXXX`

lib/payload.json
```diff
@@ -212,9 +212,19 @@
...
+                  "type": "boolean",
+                  "description": "State of Motion"
+                },
+                "motionCount": {
```
johanstokking, line 222: See above

lib/payload.json
```diff
@@ -99,6 +104,62 @@
         "air": {
           "type": "object",
           "properties": {
+            "indoor": {
```
johanstokking, line 107: I agree indoor/outdoor is super useful. And we probably should have gone with this from the start.

Now we have temperature, indoor temperature and outdoor temperature. I think instead we should add a field to indicate whether `air` was measured indoor or outdoor. This way we can reuse the same fields. Sensors able to measure indoor and outdoor concurrently can record two distinct measurements.

lib/payload.json
```diff
@@ -137,6 +198,55 @@
             }
           },
           "additionalProperties": false
+        },
+        "water": {
+          "type": "object",
+          "properties": {
+            "leak": {
+              "type": "boolean",
+              "description": "Status of water leaking (true/false)"
```
johanstokking, line 207:
```suggestion
              "description": "Leak detected"
```

This aligns with "Motion detected" below

lib/payload.json
```diff
@@ -137,6 +198,55 @@
...
+                "count": {
+                  "type": "number",
+                  "description": "Number of motion events (count)"
+                }
+              },
+              "additionalProperties": false
+            },
+            "doorState": {
+              "type": "string",
+              "description": "State of door or window",
```
johanstokking, line 231: Door and window are two separate things.

johanstokking (commented):
Great stuff already, few comments

lib/payload.json
LDannijs: These are usually reed/contact switches, used on doors or windows. not sure how to properly capture it all.

johanstokking: Then it's a more generic open/close state, don't mention door/window then.

lib/payload.json
```diff
@@ -178,9 +178,9 @@
               },
               "additionalProperties": false
             },
-            "doorState": {
+            "contactState": {
               "type": "string",
-              "description": "State of door or window",
+              "description": "State of a contact sensor",
               "enum": ["OPEN", "CLOSE"]
```
johanstokking, line 184:
```suggestion
              "enum": ["OPEN", "CLOSED"]
```

lib/payload.json
```diff
@@ -181,7 +181,7 @@
             "contactState": {
               "type": "string",
               "description": "State of a contact sensor",
-              "enum": ["OPEN", "CLOSE"]
+              "enum": ["OPEN", "CLOSED"]
```
johanstokking, line 184: And now you need to update the normalizer(s) and map the values correctly.

LDannijs: I already changed everything from `doorState` to `contactState` and there are no CLOSE states anywhere.

vendor/dragino/lsn50-v2.js
```diff
@@ -196,4 +196,21 @@ default:
...
+  return {
+    data: {
+      air: [
+        {
+          temperature: input.data.TempC_SHT,
+          relativeHumidity: input.data.Hum_SHT,
+        },
+      ],
+      action: {
+        contactState: input.data.Door_status
```
johanstokking, line 211: Regarding the above comment; you take `input.data.Door_status`. What values can it have? If we just change CLOSE to CLOSED, and leave this unchanged, something is gonna break, right?

LDannijs: Right, forgot that, apologies.

vendor/milesight-iot/ws301.js
```diff
@@ -49,7 +49,7 @@ function normalizeUplink(input) {
     return {
         data: {
             action: {
-                contactState: input.data.door.toUpperCase()
+                contactState: input.data.door ? "OPEN":"CLOSED"
```
johanstokking, line 52: So is it a boolean? Before you assumed it was a string?

johanstokking: So in this case `input.data.door` is either `close` or `open`. Not `CLOSE` or `OPEN` and not a boolean.

To convert `close` or `open` to `CLOSED` or `OPEN` respectively, you need to `contactState: input.data.door === "close" ? "CLOSED" ? input.data.door === "open ? "OPEN" : undefined`

LDannijs: I finally see how it should work now, apologies.

lib/payload.json
```diff
@@ -97,32 +102,40 @@
...
-            "relativeHumidity": {
-              "description": "Relative humidity (%)",
-              "$ref": "#/definitions/percentage"
-            },
-            "pressure": {
-              "type": "number",
-              "description": "Atmospheric pressure (hPa)",
-              "minimum": 900,
-              "maximum": 1100
+          "type": "array",
```
johanstokking, line 105: OK so I only see this now; `air` gets changed from `object` to `array`? We can't do that, that breaks API.

The solution is to return an array from the normalizer. So if there are multiple measurements, there are multiple array items, each with their own `air` object.

vendor/dragino/lwl03a.js
```diff
@@ -0,0 +1,162 @@
...
+	  data : Decoder(input.bytes, input.fPort),
+	};
+}
+
+function normalizeUplink(input) {
+  var data = {};
+  var water = {};
+
+  if (input.data.WATER_LEAK_STATUS) {
+    water.leak = input.data.WATER_LEAK_STATUS ? true : false
```
johanstokking, line 154: `WATER_LEAK_STATUS` is a string which can be `LEAK`. What value does it have when there's no leak? In any case, a ternary operator is probably not what we want here.

lib/payload.json
LDannijs: Ah i see, didnt know. Will get on it.

vendor/dragino/lwl03a.js
LDannijs: Leak status can only be `LEAK` or `NO LEAK` because of this: https://github.com/TheThingsNetwork/lorawan-devices/blob/af5ac51e9203464dfab4a15de02270750b2752d9/vendor/dragino/lwl03a.js#L124

But ive changed it so itll give false for everything that isnt `LEAK` https://github.com/TheThingsNetwork/lorawan-devices/blob/af5ac51e9203464dfab4a15de02270750b2752d9/vendor/dragino/lwl03a.js#L155

johanstokking (approved):

PR #824, opened 2024-09-18
Quandify CubicMeter 1.1 support

peter-quandify:
#### Summary
Added support for 
- CubicMeter 1.1 Plastic
- CubicMeter 1.1 Copper

#### Changes
Added files to the vendor folder for device settings and decoding

#### Checklist for Reviewers

#### Notes for Reviewers
Is the data structure returned from the decodeUplink correct?
We could not find appropriate fields in the normalizedOutput that would correspond to
- Water temperature max
- Water temperature min
- Ambient temperature
- Total volume

But we found leak status and battery level, should we add those to the normalized output object?

#### Release Notes
CubicMeter 1.1 Plastic support
CubicMeter 1.1 Copper support


LDannijs:
Hello @peter-quandify,

> Is the data structure returned from the decodeUplink correct?

Have you ran `make validate`? that will check the files if they are outputting the correct structure. I ran it on you files and it seems to be correct but if the structure wasn't correct the `make validate` would catch it.

> We could not find appropriate fields in the normalizedOutput that would correspond to
> - Water temperature max
> - Water temperature min
> - Ambient temperature
> - Total volume

> But we found leak status and battery level, should we add those to the normalized output object?

You are technically free to add your own fields to our schema, if you know how to ([documentation](https://www.thethingsindustries.com/docs/integrations/payload-formatters/javascript/uplink/#normalized-payload)). Otherwise adding as much as you can to the normalized output is recommended, as they can help out users a lot.

Jaime-Trinidad (approved):

peter-quandify:
The normalized formatter is missing the update schema definition to work properly, which we are currently working on adding.

PR #832, opened 2024-10-06
Deleted PCR2 from Vendor Parametric

akoschak:
#### Summary

...

#### Changes

Deleted PCR2 data and index.yalm

#### Checklist for Reviewers

#### Notes for Reviewers

...

#### Release Notes

- ...


Jaime-Trinidad:
Hi @akoschak why you delete the device? if the device is out of the market now, probably are users with this device, if we delete it they will stop working on TTS and will cause problems.

akoschak:
Hello Jaime

This device is now available solely from PMX Systems AG. Parametric does not support it anymore. PMX Systems (Sascha in CC) asked me to delete it

Thanks and kind Regards,
Andreas


Beste Grüsse / Kind Regards



Andreas Koschak

Owner / Managing Director | Parametric GmbH | Tel +41 33 345 01 55





________________________________

Jaime-Trinidad:
@akoschak totally understand, but still not deleting the device will be the best, as we don't know how many users are working with it, the issues will come up directly to us, it will be better if you update the `.yaml` in description and tell them the device is not supported by Parametric and now is supported by PMX system.

Jaime-Trinidad:
Hi @akoschak any comment on my last message?

sjackle:
I'm the new maintainer of the PCR2 device.
Could we set it to something like "Deprecated"? 

The profile is wrong and does not reflect the firmware anymore. So we have a lot of support.

Jaime-Trinidad:
Hi @sjackle on your side or with the files here? is there any update to do?

sjackle:
Hi Jaime

I did all the tests and waiting for acceptance of the pull request.

Please advice what's missing.
BTW: It's quite hard to do it when one does not have a real linux box and
using Windows WSL. Would it be possible to adapt the environment to also
run on WSL?

Thank you for your help.

Sascha

PR #840, opened 2024-10-30
Add Yobiiq as vendor

dhakke:
#### Summary

Added Yobiiq as vendor and the SD1001 device./

#### Changes

- ...
- ...

#### Checklist for Reviewers

- [ ] Title and description should be descriptive (Not just a serial number for example).
- [ ] `profileIDs` should not be `vendorID` and should be a unique value for every profile.
- [ ] All devices should be listed in the vendor's `index.yaml` file.
- [ ] Firmware versions can not be changed.
- [ ] At least 1 image per device and should be transparent.

#### Notes for Reviewers

...

#### Release Notes

- ...

PR #850, opened 2024-11-19
Fix profile ID for yobiiq sd-1001

KrishnaIyer:
#### Summary

Fix profile ID for yobiiq sd-1001. We missed this in https://github.com/TheThingsNetwork/lorawan-devices/pull/840 @dhakke 

#### Changes

- Change profile ID for yobiiq sd-1001 from `'04154EB7` to `20151` (`0x4EB7` in base10).  

#### Checklist for Reviewers

#### Notes for Reviewers

Vendor Profile IDs are 2 byte values as per TR005 and TTS interprets these values as base10.

#### Release Notes

- Fix profile ID for yobiiq sd-1001


KrishnaIyer:
https://github.com/TheThingsNetwork/lorawan-devices/pull/851 makes this obsolete.

PR #855, opened 2024-11-25
ezurio update

Jaime-Trinidad:
#### Summary

Laird re brand is Ezurio https://www.ezurio.com/

#### Changes

- ...
- ...

#### Checklist for Reviewers

#### Notes for Reviewers

...

#### Release Notes

- ...


johanstokking:
We need to keep all Laird devices there for backwards compatibility. Otherwise the existing fleet breaks.

PR #894, opened 2025-03-11
Refactor and Features : V1.1.2

wtc-peg:
#### Summary
Feature & Refactor: Feat: Optional batch and EP Equiv  parameters for API. Global build of v1.1.2 before new deployment for Watteco, Actility, TTN, and NPM repositories

...

#### Changes

- Feature: Add decodeUplink API Optional parameters allowing decoding customization by user when Batch config or Endpoint equivalence changed
- Refactor namming, json device files and version management for actility deployment
- Refactor Array in examples YAML files during TTN deployment

#### Checklist for Reviewers

#### Notes for Reviewers

...

#### Release Notes

- ...


Jaime-Trinidad:
Hi @wtc-peg there is an inconsistency in the decoder output related to the DebP field. The value of this field differs depending on the environment where the decoder is executed (e.g., local machine vs. GitHub CI), which causes the validation to fail.

The issue appears to come from the handling of timestamps or date conversion in the `normalisation_batch` function, located in `vendor/watteco/ticso.js`

Please review the `DebP` field and ensure that the date output is consistent across all environments, regardless of the local timezone.

PR #899, opened 2025-03-18
Add tehama pulse-sensor

MrJohnny:
#### Summary

...

#### Changes

- ...
- ...

#### Checklist for Reviewers

#### Notes for Reviewers

...

#### Release Notes

- ...


Jaime-Trinidad:
Hi @MrJohnny thank you for your PR, please sign the CLA

MrJohnny:
Jaime-Trinidad, I've clicked "CLA not signed" button twice and "successfully" signed it... at least that was what the webpage said. I think there is a problem with my having committed these changes as "root" rather than as "MrJohnny"

MrJohnny:
@Jaime-Trinidad, I've clicked "CLA not signed" button twice and "successfully" signed it... at least that was what the webpage said. I think there is a problem with my having committed these changes as "root" rather than as "MrJohnny"

PR #923, opened 2025-06-01
Decoder update to v2.2.1. VoBoSL removal

mvujovic-vb:
#### Summary

Decoder update to v2.2.1. VoBoSL removal.

#### Changes

- Decoder is updated to v2.2.1
- VoBoSL is removed

#### Checklist for Reviewers

#### Notes for Reviewers

...

#### Release Notes

- ...


johanstokking (approved):
LGTM, one question though

PR #978, opened 2025-11-21
add milesight encoder

harryhua-ai:
#### Summary

...

#### Changes

- ...
- ...

#### Checklist for Reviewers

- [x] Title and description should be descriptive (Not just a serial number for example).
- [x] `profileIDs` should not be `vendorID` and should be a unique value for every profile.
- [x] All devices should be listed in the vendor's `index.yaml` file.
- [x] Firmware versions can not be changed.
- [x] At least 1 image per device and should be transparent.

#### Notes for Reviewers

...

#### Release Notes

- ...


harryhua-ai:
@johanstokking  hello Johan

I add the encoder of milesight thermalstat. Just a quick question:  is it possible to spit the wt201v2.js to wt201v2.js and wt201v2-encoder.js? 

I have tried a test with wt201v2.js and wt201v2-encoder.js, then link it in wt201v2-codec.yaml with the correct file
```
uplinkDecoder:
  fileName: wt201v2.js

downlinkEncoder:
  fileName: wt201v2-encoder.js
```
the validate process works perfectly. 

can I do it like this? 
currently, both the decoder and encoder scripts are merged in the same file.

Split it into 2 files will speed up my PR for more models.

harryhua-ai:
> > I add the encoder of milesight thermalstat. Just a quick question: is it possible to spit the wt201v2.js to wt201v2.js and wt201v2-encoder.js?
> 
> Yes, you can do it.
> 
> If you define examples like in https://github.com/TheThingsNetwork/lorawan-devices/?tab=readme-ov-file#payload-codecs, you're certain it works.

nice, thanks :)

PR #988, opened 2026-01-08
Edit inBiot Devices to fit Market definition and new Firmware

idoate:
#### Summary

This pull request EDIT the initial set of inBiot device definitions to the TTN LoRaWAN Device Repository.

It includes the basic files required for proper device onboarding, following the repository structure and validation rules.
#### Changes

- EDIT vendor folder structure for inBiot.

- Included index.yaml listing all submitted devices.

- EDIT individual device profiles and metadata files (basic definitions).

- EDIT device images (transparent PNG format).

- Ensured profileIDs are unique and follow TTN conventions.

#### Checklist for Reviewers

- [ ] Title and description should be descriptive (Not just a serial number for example).
- [ ] `profileIDs` should not be `vendorID` and should be a unique value for every profile.
- [ ] All devices should be listed in the vendor's `index.yaml` file.
- [ ] Firmware versions can not be changed.
- [ ] At least 1 image per device and should be transparent.

#### Notes for Reviewers

...

#### Release Notes

-EDIT initial device definitions and vendor structure for inBiot in the LoRaWAN Device Repository.

- Included basic metadata and transparent images for supported devices.

PR #1031, opened 2026-06-24
Inbiot PR 988

Jaime-Trinidad:
#### Summary
This pull request replace #988 . Edit the initial set of inBiot device definitions to the TTN LoRaWAN Device Repository.

It includes the basic files required for proper device onboarding, following the repository structure and validation rules.

...

#### Changes

- EDIT vendor folder structure for inBiot.
- Included index.yaml listing all submitted devices.
- EDIT individual device profiles and metadata files (basic definitions).
- EDIT device images (transparent PNG format).
- Ensured profileIDs are unique and follow TTN conventions.

#### Checklist for Reviewers

#### Notes for Reviewers

...

#### Release Notes

- ...


vendor/inbiot/mini-lora.yaml
```diff
@@ -8,7 +8,7 @@ hardwareVersions:
     numeric: 2
 
 firmwareVersions:
-  - version: '2.1'
+  - version: '2.X'
     numeric: 1
```
johanstokking, line 12: This is risky as it may break existing devices.

Since there's quite a bit of removal going on here, are existing devices a concern?

idoate: We've included this version because it can be updated via OTA without affecting LoRa functionality.

Currently, the devices are running version 2.84.

Would you prefer that we include this even though it might become outdated, or should we leave the 2.x version as a a general reference?

johanstokking: In that case I would suggest adding a new firmware version that is `2.x`, and leave `2.1` unchanged. Because end-devices registered with `2.1` will otherwise lose their codec.

idoate: @johanstokking Thanks for the feedback we've addressed it as suggested: the 2.1 firmware version is restored unchanged (so end-devices registered with 2.1 keep their codec), and 2.X has been added as a new firmware version entry (numeric: 2) in all four device files (mini-lora.yaml, mica-lora.yaml, plus-lora.yaml, well-lora.yaml). The changes pass npm run validate locally.

Regarding the removals: the -mdb variants are Modbus-only configurations that were listed in this repository by mistake their removal is intentional and there are no LoRaWAN end-devices registered under those IDs.

@Jaime-Trinidad  the fix is ready in  https://github.com/Jaime-Trinidad/lorawan-devices/pull/1  targeting your inbiot-update branch please review and merge it so this PR picks up the commit. Thanks!

johanstokking (approved):

vendor/inbiot/mini-lora.yaml
johanstokking: Thanks @idoate for the update and the clarification. I was able to append the commit to the PR branch here.

PR #1056, opened 2026-08-20
Build the website in CI, scoped away from vendor contributions

wienke:
**Updated after review feedback from @mjamescompton — the original version of this PR built the website on every pull request, which was the wrong design. See "Scoping" below.**

`Validate` runs only `npm run validate`, which never touches the Hugo site or the webpack bundle. Anything that breaks the website build passes CI and surfaces later in `Release Device Repository`, after the merge has already landed on master.

This is not theoretical: the four `/website` security bumps merged today — postcss, fast-uri, webpack-dev-server and shell-quote — had no CI coverage at all. They were built locally before merging, but nothing in the repo would have stopped a bad one. The failure mode is at least safe rather than destructive (if the build breaks, the deploy step never runs and `gh-pages` keeps serving the last good copy) but master is left in a state where every subsequent release fails.

### Scoping — why this is not a check on every PR

This repository is a source of **device data that vendors contribute to**, and the website is a downstream consumer of that data, not a gate on it. It is not a vendor's responsibility to keep the Hugo site building.

Around 40% of recent merged pull requests touch only `vendor/`, opened by device manufacturers. Showing them a red X caused by an internal Hugo or webpack problem would be confusing and unfair, and would make an internal website breakage look like a blocker on their device submission.

So this is a **separate workflow, scoped to the files the website build actually reads**:

- `website/**`
- `package.json` / `package-lock.json` — consumed by the `npm ci` step
- `schema.json` and `lib/**` — `website/tools/compile-schema.js` compiles these into standalone validators

A pull request touching only `vendor/` never triggers it. **`Validate` is left exactly as it was**, so nothing changes for vendor contributions.

Vendor data reaching the site stays covered where it is covered today: the generator consumes `vendor/` during `Release Device Repository` on every push to master.

### What runs

Nine build steps, **identical** to the build half of the `pages` job in `release.yml` — same Go (`~1.24`), Node (22), Hugo (0.159.2 extended), same `BASE_URL` / `BASE_PATH` — stopping immediately before the deploy, `devices.csv` and Airtable steps. A green check means the release build will succeed. Verified end-to-end: all nine steps pass.

### One caveat before making it required

A required status check that is **skipped by a paths filter never reports**, and GitHub then blocks the pull request while it waits for it. Making this check required as-is would block every vendor-only PR indefinitely — precisely the outcome the scoping above is meant to avoid.

If it should eventually be required, the usual pattern is a companion job with the inverse paths filter that reports success, so the context always reports one way or the other. Happy to add that in a follow-up; leaving it non-required for now is the safer default.

### DIFF
--- a/.github/workflows/build-website.yml
+++ b/.github/workflows/build-website.yml
@@ -16,16 +16,12 @@ on:
       - master
     paths:
       - 'website/**'
-      - 'package.json'
-      - 'package-lock.json'
       - 'schema.json'
       - 'lib/**'
       - '.github/workflows/build-website.yml'
   pull_request:
     paths:
       - 'website/**'
-      - 'package.json'
-      - 'package-lock.json'
       - 'schema.json'
       - 'lib/**'
       - '.github/workflows/build-website.yml'
@@ -54,8 +50,6 @@ jobs:
         run: |
           cd website/
           make go.deps
-      - name: Install Node.js dependencies
-        run: npm ci
       - name: Build content
         run: |
           cd website/
--- a/.github/workflows/build-website.yml
+++ b/.github/workflows/build-website.yml
@@ -1,0 +1,77 @@
+# The website is a downstream consumer of the device data in this repository,
+# not a gate on it: vendors contribute device YAML, codecs and images, and it
+# is not their responsibility to keep the Hugo site building. This workflow is
+# therefore scoped to the files the website build actually reads, so a pull
+# request that only touches vendor/ never triggers it -- an internal website
+# problem can never show up as a failure on a vendor's device submission.
+#
+# Vendor data reaching the site is still covered: the generator consumes
+# vendor/ during Release Device Repository on every push to master, which is
+# where such a break surfaces today.
+name: Build website
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - 'website/**'
+      - 'package.json'
+      - 'package-lock.json'
+      - 'schema.json'
+      - 'lib/**'
+      - '.github/workflows/build-website.yml'
+  pull_request:
+    paths:
+      - 'website/**'
+      - 'package.json'
+      - 'package-lock.json'
+      - 'schema.json'
+      - 'lib/**'
+      - '.github/workflows/build-website.yml'
+
+jobs:
+  build-website:
+    name: Build website
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v4
+      - name: Set up Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: '~1.24'
+      - name: Set up Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: '22'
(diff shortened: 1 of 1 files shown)


mjamescompton:
I understand the idea but

Its not our Vendors responsibility to fix the website. 

This would block our Vendors from updating the device repository if we had an internal problem with the website.

.github/workflows/build-website.yml
```diff
@@ -0,0 +1,77 @@
...
+      - 'package.json'
+      - 'package-lock.json'
+      - 'schema.json'
+      - 'lib/**'
+      - '.github/workflows/build-website.yml'
+  pull_request:
+    paths:
+      - 'website/**'
+      - 'package.json'
+      - 'package-lock.json'
```
mjamescompton: see above

.github/workflows/build-website.yml
```diff
@@ -0,0 +1,77 @@
...
+name: Build website
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - 'website/**'
+      - 'package.json'
+      - 'package-lock.json'
```
mjamescompton: The website maintainer should not be concerned with the `package.json` of the cli tools. There is a second `package.json` in the website/ directory

mjamescompton (changes requested):
The goal of this PR is not quite clear to me, do you want to validate that the website will always build after any update to the code of the website or of any part of the device repository. Is this meant to be a guard for the dependabot PRs?

wienke:
Both points addressed in a0de5a9. You're right about the root package.json — the full build chain (go.build, yarn build including the prebuild schema compile, hugo.build.public) builds cleanly with root node_modules absent; website/package.json declares its own ajv/ajv-formats. The root npm ci in release.yml exists for the later 'make devices.csv' step, which this workflow stops before. Dropped the step and the root package.json/lockfile paths.

On the goal: the motivating case was the dependabot PRs (the four /website bumps merged today had zero CI coverage), but it's not only a dependabot guard — it covers any PR touching what the website build consumes: website/ itself, and schema.json + lib/ which compile-schema.js compiles into the wizard's validators. Deliberately NOT vendor/ or the rest of the repo: vendor data reaching the site is already covered by the generator run in Release on every master push, and an internal build problem should never gate a vendor submission.

wienke:
To answer your question directly: yes — the trigger was the dependabot PRs. Yesterday's eight security merges included four /website bumps, and nothing in CI builds the website, so they went in unverified. This workflow closes that gap for the files the website build actually reads, and deliberately nothing else — vendor PRs never trigger it.

That said, whether we want this workflow at all is a maintainer call, and .github/ is Johan's area — so no rush and no re-review needed now. I'm parking this until he's back. If you two would rather rely on the release pipeline failing safe on master, closing this is fine by me.

(One last commit while parking: 599ffb3 restricts the workflow token to contents: read, since the job runs PR-supplied build code.)

johanstokking (approved):

mjamescompton (approved):

